PR #1, opened 2019-08-26
Add APIcast SelfManaged implementation

miguelsorianod:
This PR implements the APIcast custom resource to be able to deploy APIcast self-managed in a standalone way without the full 3scale infrastructure.

One notable change in comparison with the apicast self-managed templates that we provide is that this operators is Kubernetes-native, allowing us to deploy in Kubernetes OperatorHub.

The APIcast CR structure is totally different and independent than the included APIcasts in the whole 3scale api management infrastructure deployed through 3scale-operator (https://github.com/3scale/3scale-operator)

Some remarks on the decisions taken and the motivations of them:

### Image Management

Due to we want to be Kubernetes-native we cannot use ImageStreams anymore.
The way  to reference an image in Kubernetes-Native environment is by specifying the desired docker image URL reference in the Deployment where the image is referenced.
A policy on how the image is pulled can be configured through the ImagePullPolicy of the Deployment kind (https://kubernetes.io/docs/concepts/containers/images/)
Possible values for ImagePullPolicy are:
* Always: means that kubelet always attempts to pull the latest image. Container will fail If the pull fails
* Never: means that kubelet never pulls an image, but only uses a local image. Container will fail if the image isn't present
* IfNotPresent: means that kubelet pulls if the image isn't present on disk. Container will fail if the image isn't present and the pull fails

None of the available options really fits with the use case where the referenced image is a docker image floating tag, which is the way we have been managing APIcast releases until now when providing apicast self-managed.
This seems to be confirmed by the fact that the Kubernetes documentation says that "latest" tag (a floating tag) should not be used https://kubernetes.io/docs/concepts/containers/images/#updating-images

We don't want to use "Always" as the pull policy because new versions of the image would be pulled in the case where APIcast pod would be redeployed or scaled up. This would make the obtainment of the image non-controlable which we don't want

We don't want to use "IfNotPresent" because in case we were to update the image floating tag then we would never obtain the new image because it would be already present on disk

We don't want to use "Never" as never obtaining the image would mean we could never update it.

The only option that I can think of right now is using a fixed tag image, and release a new operator version each time there's a new APIcast self-managed image. The downside of that approach is that even for CVE image rebuilds we would have to release a new operator version (also for bugfixes and new 3scale releases in apicast-selfmanaged).

### Secrets and Configuration management

Now APIcast is deployed through a Deployment and not through a DeploymentConfig(only available in OpenShift).
In order to try to make application-level configuration in APIcast flexible and not tied to the operator code, a secret can be provided via the APIcast CR that contains the environment variables that the APIcast pod will contain. NO specific environment variable names are specified in the code and no defaults are set either for the pods.
This secret is then referenced in the Deployment with the "EnvFrom" attribute. The downside in that approach is that changes in the Secret do NOT trigger changes in the Deployment. This behavior is a limitation/behavior is native to Kubernetes. This would mean that when changing a secret that is referenced in the Deployment the changes would not be effective in the pods of the deployment unless the user manually redeploys the pods of the Deployment.

To try to solve this limitation imposed by Kubernetes, the following approach has been taken:
 * The PodTemplate annotations section of the Deployment is updated with the resourceVersion of the secret that uses
 * In order to detect changes on the Secret provided via the CR an OwnerReference is set to them by the Operator and also a Watch of the secrets owned by the operator has been added. This also means that even though the secret is originally created by the user, it will be removed when removing the APIcast related CR.
 * When a change is detected in the referenced Secret by comparing the existing Secret resourceVersion and the saved one in the PodTemplate annotation section then this annotation is updated, triggering a new redeploy of the Deployment pods making the changes effective

This also applies to the 'apicast-<crName>-admin-portal-endpoint' secret that is only managed by the Operator and it is not intended to be edited by the users.
Also other watches and controllerreferences are added on other secrets that have to be provided in the CR as fields.

### Deployment Management

OpenShift's DeploymentConfig's allowed the redeployment of the pods at least through the UI. With Deployments it seems it is not possible to manually redeploy a Deployment (https://kubernetes.io/docs/concepts/workloads/controllers/deployment/).
If for some reason a user wants to redeploy pods in a Deployment without having really changed anything in the deployment then the solution would be for example adding an arbitrary "dummy" annotation in the PodTemplate to force Kubernetes detecting a change that redeploys it.

### Operands resource names management

In order to try to make multiple instances of an APIcast CR available in the same project, all the resources created by the APIcast operator when deploying a CR will have the "apicast-" prefix. For example, if the APIcast CR you deploy has the name "myapicast" then the Deployment that is created by the operator will have the name "apicast-myapicast"

### Upgrade Management

For potential changes that might appear in the future in the Operator that we don't want to be reconciled an upgrade procedure is coded as part of the Operator logic that will have to contain the needed changes for every operator release that needs them. This means that new code will for the upgrade will have to be implemented each time there are this kind of changes.

For the moment there's logic that sets the current operator version that the CR belongs to via an annotation named "apicast.apps.3scale.net/operator-version".
In case the annotation is not set in the CR the operator will set it. In case it is already set in the Operator and it does not match the deployed operator version the upgrade procedure will be triggered and when it is finished the annotation will be updated with the existing deployed operator version

### Public hostname exposure

There's a field in the APIcast CR named "ExposedHostname" that allows making possible the ability to make the APIcast endpoint available from outside the cluster.
The way it does this is by creating an Ingress associated to the underlying APIcast's Kubernetes Service.
Take into account that creating the service does not always necessarily means that the endpoint is accessible automatically. Frequently it requires that the user will have to create and configures the desired IngressController themselves before/after deploying APIcast, which is a totally different action and not responsibility of this operator. The operator is responsible to provide the ingress so the endpoint can be made public, but how it is made public it is up to the user depending on their needs.
You can read more about the relations between Ingress and Ingress Operator here https://kubernetes.io/docs/concepts/services-networking/ingress/#what-is-ingress and some of the available Ingress Controllers that are available here https://kubernetes.io/docs/concepts/services-networking/ingress-controllers/

Also it has to be taken into account that if the user decides to configure TLS at the Ingress level only the port 443 is configured an TLS termination is assumed. This is a limitation of Kubernetes Ingress objects themselves. More information here https://kubernetes.io/docs/concepts/services-networking/ingress/#tls

### Reconciliation

Only fields directly referenced by the CR (or that have a side-effect from them) are reconciled.

### Status section management

No fields in status have been implemented yet. If we deem that necessary we'll implement them in another PR


miguelsorianod:
I add an example of an APIcast CR:

```
kind: APIcast
metadata:
  name: example-apicast
spec:
  replicas: 1
  adminPortal:
    urlSecretKeyRef:
      name: apicast-admin-portal
      key: url
    accessTokenSecretKeyRef:
      name: apicast-admin-portal
      key: accesstoken
  environmentConfigurationSecretRef:
    name: apicast-environment
  image: quay.io/3scale/apicast:nightly
```

The apicast-admin-portal and apicast-environment secrets must have been created before and the apicast-admin-portal secret with the specified keys.

Another example with an exposed hostname as an Ingress would be:

```
kind: APIcast
metadata:
  name: example-apicast
spec:
  replicas: 1
  adminPortal:
    urlSecretKeyRef:
      name: apicast-admin-portal
      key: url
    accessTokenSecretKeyRef:
      name: apicast-admin-portal
      key: accesstoken
  environmentConfigurationSecretRef:
    name: apicast-environment
  image: quay.io/3scale/apicast:nightly
  exposedHost:
    host: "myhost.com"
```

eguzki:
First of all. good job :medal_military: 

My general comments before reviewing the code, based on design decision exposed in this PR:

Image management
--------------------------
Fully agree on following k8s best practices and avoiding use of floating tags like latest. Indeed, secfixes and bugfixes will require new operator releases.

Good you exposed image in the CRD. However, as well as 3scale Operator, we have to make it clear that it is only meant for devtesting. Users should not use image field in the CR or would miss secfixes, bugfixes or even upgrading capabilities.

Secrets and Configuration management
---------------------------------------------------
I have serious doubts about the implemented solution. I guess the motivation is to make some configuration (not all, like admin portal url) operator code independent. The tradeoff is a complicated (_complicated_ always comes along with error prone) mechanism based on watching a secret and detecting secret changes to redeploy a deployment based on a `trick` which is updating annotation. 

What I suggest: KISS (keep it short and simple). Why not include them in the CR and propagate as env values?? 
```yaml
env:
    - name: DEMO_GREETING
      value: cr.demo_greeting
    - name: DEMO_FAREWELL
      value: cr.demo_farewell
```

CR is a great interface to expose configuration parameters that the user eventually will want to update (and to be reconciled). Exceptions to that general rule? Credentials or sensitive information (in secrets) or configuration params not supported in reconciliation (in config maps??). I would start thinking about which configuration parameters are we going to support for reconciliation? which ones are left out of reconciliation?

You get out-of-the-box redeployment of the operand. The downside? I do not think it is at level of operator source code maintenance. I would think about CRD's would change adding new params. But that is expected too. Removing params or updating param names will break backwards compatibility anyway we implement.

I highly recommend rewriting this first implementation as simple as possible. 

Operands resource names management
-----------------------------------------------------
Good about namespacing :+1: . It is likely a user will want to deploy 2 CR's, one for staging and another for production.

 Upgrade Management
------------------------------
I would leave this for the moment. We can leave the code as it is, but we can review it later.

Reconciliation
------------------
We are still learning about this complex procedure. So far, the best and most flexible implementation approach is that one you propose: reconciliation on k8s resource fields which value is directly or indirectly product from CR fields.

miguelsorianod:
Hi @eguzki ,

Thank you for your feedback.

Regarding your comment about

> Secrets and Configuration management

using CR fields instead of ConfigMap/Secret, I don't really see the Secret approach that bad.
I'll try to say why:

The motivation for that is to have the operator code independent of the environment variable names as much as possible indeed.

> I guess the motivation is to make some configuration (not all, like admin portal url) 

Admin portal URL has to be taken into account separately due to this information is mandatory, is a Secret (should not be provided in the CR), cannot be known in advance by the operator and in my opinion is "conceptually" different than the application-level environment variables (the url could be provided by another operator etc.).

> The tradeoff is a complicated (complicated always comes along with error prone) mechanism based on watching a secret and detecting secret changes to redeploy a deployment based on a trick which is updating annotation.

I don't think this is that uncommon:
 * A common pattern in an Operator, when you are interested on acting upon changes on resources that are managed by the Operator is to add a Watch into those resource types. Those are called secondary resources. You have an example in the operator-sdk repository itself https://github.com/operator-framework/operator-sdk-samples/blob/master/memcached-operator/pkg/controller/memcached/memcached_controller.go#L54 although you'll observe that in several operator repositories
 * The pattern of keeping an account of some specific "version" or state of an object to act upon changes on it doesn't seem to be uncommon either. For example, Kubernetes itself in its APIConventions https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md says:
  > Some resources report the observedGeneration, which is the generation most recently observed by the component responsible for acting upon changes to the desired state of the resource. This can be used, for instance, to ensure that the reported status reflects the most recent desired status.

  You can see an example of this for example in the Deployment resource types, and I also have seen it in other projects.
  I've also seen other approaches that try to keep note of the changes on a managed resource. For example the openshift registry operator saves a checksum of them in an annotation https://github.com/openshift/cluster-image-registry-operator/blob/5a6d42a35f978671c21d65fb95305952bb036ab8/pkg/resource/deployment.go#L83
and I've also seen the storage of the checksum of the spec itself too in some other project.
 * The pattern of making a change in a Deployment by changing the PodTemplateSpec does not seem to be really uncommon neither, as there's no other mechanism provided in Deployments and this might be desirable anyway. See previous link where they do it too for example. I've also found another example where in Kubernetes 1.15 they've implemented a "kubectl rollout restart" command to automatically rollout a new deployment and the way they do it is by adding an annotation too https://github.com/kubernetes/kubernetes/pull/76062/files so it's not really that uncommon.

I don't really see it as that bad of an option.

We also really seem to need an implementation of this at least for the admin-portal part, as it is provided as a secret. How else would you do that part?

I agree that the approach is not perfect but that seems what we have available with how Kubernetes Deployment works and references to Secrets/ConfigMaps and reaction to object changes work. Of course there might be bugs related to this (as with any other functionality) and if it causes big problems we would have to reconsider it, but even in that case at this moment I don't see a good solution for when you have sensitive data as we shouldn't specify the secrets directly in the CR.

About being directly an environment section as an array of environment variables, my reservations about that approach are:
 * If you look at it, the environment in the CR I implemented is a Secret and not a ConfigMap. We might have environment variables that are sensitive. In that case the best thing is that they are referenced by a secret as we don't want to specify the sensitive data in the CR itself. If they are referenced by a secret you still need to know when the secret changes and restart the deployment upon it.
 * If some environment variable is not inherently reconcilable out of the box (it could happen for example if the environment variable is "deploy-only" or implies changes in the infrastructure colaterally) I thin the problem would happen in both the Secret and direct CR env specification approach. I don't think there's a difference in that case. It should be documented accordingly in that case for both cases that is non-reconcilable if you specify it, and if it requires referencing that specific envvar name in the code to give a special treatment to it, it would be needed in both cases.
 * Once you have an array you have to begin taking into account the order of it throughout your code and merging of different sections of the CR that end up in the same environment section.

eguzki:
CR is the primary interface exposed to the user to trigger updates you want the operator perform. I think anything you allow the user to change in the deployment configuration, should be exposed in the CR. infraestructure level attributes like "replica" count and "application" level attributes like some environment variable, becase they are all part of deployment config or service or imagestream you want to control. I disagree making that differentiation between infrastructure and application level params. Application level params are part of the deployment and you want the operator to have control over them. Validation, specific procedures and so on in operator logic. 

You do not want the user to manually update your owned k8s resources. You want the operator to do that. Therefore, expose them in CR. Exceptions? As I said before, of course, sensitive data. Of course access token should not go in CR. And you are right saying we still need to watch on secrets if we allow reconciliation of those sensitive data. For instance, if we allow changing accesstoken, the secret that accesstoken exist must be watched and information propagated to the deployment config or whatever. The mechanism you implemented seems to be valid (let's see it in action...) and I agree it should be used to reconcile secrets. What I do not see is putting in secrets parameters exposed to the user not sensitive. The CR was meant for those parameters. Using secrets you would lose CR (openapi) validation, extensions being developed in OCP to edit CR's yamls with a better user experience UI, etc...

Therefore, attributes exposed to the customer:

* Sensitive data that needs to be reconciled: in secrets and secrets are watched and reconciled with the mechanism implemented.
* Sensitive data install only:  in secrets and secrets are not watched
* Attributes that need to be reconciled: in CR. CR fields that are env variables in deployment configs, are copied as they are, so reconciliation is easy and straightforward
* Attributes install only: in config maps

Specifically to our use case:

* Admin Portal URL and access token should be reconciled and they should exist in the same secret. Otherwise reconciliation would happen in more than one step and it would be nice to avoid that transitive period we are creating with two secrets.
* List all parameters that can be reconciled and put them in the CR, like `THREESCALE_DEPLOYMENT_ENV`
* Leave the rest of the parameters install only y some config map. Are there any? Since there is no storage, maybe everything should be able to reconcile.
* I am not sure Admin Portal URL should be mandatory field. We do not want to allow deploying apicast with `THREESCALE_CONFIG_FILE`?? Why not? The customer would like to create configmap with the config file and the deployment runs apicast without `THREESCALE_PORTAL_ENDPOINT`.

miguelsorianod:
Regarding

> Admin Portal URL and access token should be reconciled and they should exist in the same secret. Otherwise reconciliation would happen in more than one step and it would be nice to avoid that transitive period we are creating with two secrets.

Ok. I can't see a specific reason right now that mandates having them separated.
I've pushed a new commit that contains this change. At the end of the PR I'll redo history to squash it.

miguelsorianod:
Regarding

> List all parameters that can be reconciled and put them in the CR, like THREESCALE_DEPLOYMENT_ENV

By doing so instead of placing all of them into all of them in a secret this has the following implications:
 * Each time we want to support a new APIcast deployment envvar we'll have to add new changes in the CRD. This also means that we will have to upgrade the CRD
 * We risk being in a situation where we realize we might want to reorganize some set of parameters in a different section. This implies a change in a CRD that would break compatibility, which normal upgrade processes cannot perform and do not seem to be supported right now in operator-sdk in an automated way (it seems one of the solutions is using webhooks which are currently not supported there)

I understand your point of view where it has the following favourable points:
 * The user does not need to create a separate secret for fields that do not contain sensitive data, having them in the CR is more comfortable. This point however implies that for sensitive data we might end up with more than 1 secret for that sensitive data instead of the originally required 1 secret only for them.
 * We can take advantage of OpenAPI validation

I'm not entirely convinced that this is a better solution than the proposed one using secrets, but let's give it a try... I'll make the needed changes.

miguelsorianod:
Regarding
> I am not sure Admin Portal URL should be mandatory field. We do not want to allow deploying apicast with THREESCALE_CONFIG_FILE?? Why not? The customer would like to create configmap with the config file and the deployment runs apicast without THREESCALE_PORTAL_ENDPOINT.

I already had a PR in progress that would take care of that. I'll work on that in another PR though so lets exclude this point from this PR. That change will make us decide whether the Admin URL should be mandatory or not. There are several tradeoffs that we will have to end discussing.

miguelsorianod:
Regarding,

> Leave the rest of the parameters install only y some config map. Are there any? Since there is no storage, maybe everything should be able to reconcile.

At this moment I don't know.
With the new CR inclusion approach of the fields I'll look what I find. I'll begin by configuring the same fields that the template offers for the envvars and see what I find

miguelsorianod:
I've replaced the AdditionalEnvironment secret to direct CR spec fields.
I've added the ones that are provided with the existing OpenShift template that we have excluding OAuth2 which are in process to be deprecated.

There is some commented code that will be cleant up but before doing that I'd like that we review the current state of the PR.

pkg/apis/apps/v1alpha1/apicast_types.go
```diff
@@ -13,14 +15,95 @@ type APIcastSpec struct {
...
+	EnvironmentConfigurationSecretRef *v1.SecretEnvSource `json:"environmentConfigurationSecretRef,omitempty"`
+	// +optional
+	ServiceAccount *string `json:"serviceAccount,omitempty"`
+	// +optional
+	Image *string `json:"image,omitempty"`
+	// +optional
+	ExposedHost *APIcastExposedHost `json:"exposedHost,omitempty"`
+
+	// +optional
+	DeploymentEnvironment *DeploymentEnvironmentType `json:"deploymentEnvironment,omitempty"` // THREESCALE_DEPLOYMENT_ENV
```
miguelsorianod: I tried to define a type for the attributes that only accept a set of values but it seems that operator-sdk generator does not autogenerate openapi specs that limit to those value so I'm not really sure whether this is worth it, as it has some consequences (tying to a type, having to convert to string sometimes etc.)

pkg/apicast/apicast.go
```diff
@@ -0,0 +1,357 @@
...
+
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
+	extensions "k8s.io/api/extensions/v1beta1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+type APIcast struct {
```
miguelsorianod: The idea with this own type, is to have an "internal" type independent of the CRD type so they are independent to changes.

pkg/apicast/apicast.go
```diff
@@ -0,0 +1,357 @@
...
+	ServiceName            string
+	Replicas               int32
+	AppLabel               string
+	ServiceAccountName     string
+	Image                  string
+	ExposedHost            ExposedHost
+	OwnerReference         *metav1.OwnerReference
+	AdminPortalCredentials AdminPortalCredentials
+
+	DeploymentEnvironment          *string
```
miguelsorianod: Here I've used string instead of the custom defined type in the apicast_types.go as I don't want to be dependent on "external" types.

pkg/controller/apicast/apicast_logic_reconciler.go
```diff
@@ -0,0 +1,459 @@
...
+		changed = true
+
+	}
+	if !reflect.DeepEqual(existingDeployment.Spec.Template.Spec.ServiceAccountName, desiredDeployment.Spec.Template.Spec.ServiceAccountName) {
+		changed = true
+		existingDeployment.Spec.Template.Spec.ServiceAccountName = desiredDeployment.Spec.Template.Spec.ServiceAccountName
+
+	}
+
+	if !reflect.DeepEqual(existingDeployment.Spec.Template.Spec.Containers[0].Env, desiredDeployment.Spec.Template.Spec.Containers[0].Env) {
```
miguelsorianod: Not sure whether we should first sort both arrays before performing the comparison

eguzki:
I do not see the reconciliation logic for the secret having info about admin portal url. 

If it is not implemented yet, I suggest we leave it for another PR to avoid making this PR too big.

deploy/crds/apps_v1alpha1_apicast_crd.yaml
```diff
@@ -28,8 +28,90 @@ spec:
...
+              type: integer
+            deploymentEnvironment:
+              type: string
+            dnsResolverAddress:
+              type: string
+            enabledServices:
+              items:
+                type: string
+              type: array
+            environmentConfigurationSecretRef:
```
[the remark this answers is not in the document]
miguelsorianod: No, this is wrong. It's because the attribute was not removed from the APIcastSpec type. I'll remove it :+1:

pkg/controller/apicast/apicast_controller.go
```diff
@@ -51,9 +72,32 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+
+	err = c.Watch(&source.Kind{Type: &appsv1.Deployment{}}, &handler.EnqueueRequestForOwner{
+		IsController: true,
+		OwnerType:    &appsv1alpha1.APIcast{},
+	})
+	if err != nil {
+		return err
+	}
+
+	err = c.Watch(&source.Kind{Type: &v1.Service{}}, &handler.EnqueueRequestForOwner{
```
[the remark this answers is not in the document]
miguelsorianod: The idea is to generate events for the objects that are managed by the APIManager so we can detect changes in their state and re-reconcile them to the desired state.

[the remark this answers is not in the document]
miguelsorianod: In the specific case of Ingres, The rules (host name is involved there) and the TLS config.
Basically the section named "exposedHost" in the CR

pkg/controller/apicast/apicast_controller.go
```diff
@@ -86,68 +131,82 @@ func (r *ReconcileAPIcast) Reconcile(request reconcile.Request) (reconcile.Resul
...
-	if err != nil {
-		if errors.IsNotFound(err) {
-			// Request object not found, could have been deleted after reconcile request.
-			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
-			// Return and don't requeue
-			return reconcile.Result{}, nil
-		}
-		// Error reading the object - requeue the request.
+	instance, err := r.getAPIcast(request)
+	if instance == nil {
```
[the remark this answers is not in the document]
miguelsorianod: In case instance is nil it means something has gone wrong and we return the error appropiately.
When the error is != nil the request is requeued.

miguelsorianod:
Regarding the reconciliation logic for the secret having info about admin portal url, you should see the code in the "reconcileAdminPortalCredentials" method in apicast_logic_reconciler.go

Also, in the reconcileDeployment method you should see the reconciliation of the corresponding annotation

deploy/crds/apps_v1alpha1_apicast_crd.yaml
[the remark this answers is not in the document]
miguelsorianod: Fixed

pkg/controller/apicast/apicast_controller.go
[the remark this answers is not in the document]
eguzki: this is not very conventional. 

Test error to react upon error. Maybe just logging?

Test instance to return when it does not exist.

Leave it like that if you want. For me this is ugly code.

eguzki:
you need to rebase before merge

miguelsorianod:
I changed the unconventional error checking and rebased commits.
Ready for review

pkg/controller/apicast/apicast_logic_reconciler.go
```diff
@@ -0,0 +1,431 @@
...
+}
+
+func (r *APIcastLogicReconciler) namespacedName(object metav1.Object) types.NamespacedName {
+	return types.NamespacedName{
+		Name:      object.GetName(),
+		Namespace: object.GetNamespace(),
+	}
+}
+
+func (r *APIcastLogicReconciler) getAPIcast(request reconcile.Request) (*appsv1alpha1.APIcast, error) {
```
eguzki: used?

pkg/controller/apicast/apicast_logic_reconciler.go
```diff
@@ -0,0 +1,431 @@
...
+	desiredAPIcast, err := r.internalAPIcast()
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	desiredAPIcastAdminPortalEndpointSecret, err := desiredAPIcast.AdminPortalEndpointSecret()
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	err = r.reconcileAdminPortalEndpointSecret(desiredAPIcastAdminPortalEndpointSecret)
```
eguzki: reconcile admin portal secret was not done in "internalAPICast` method. This is confusing

pkg/controller/apicast/apicast_logic_reconciler.go
```diff
@@ -0,0 +1,431 @@
...
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	adminPortalEndpointSecret, err := desiredAPIcast.AdminPortalEndpointSecret()
```
eguzki: second call to AdminPortalEndpointSecret() to create desired secret.

What I would do:

1.- Read secret. If it does not exist, return empty secret.
2.- Create structure with info required to build the secret. If a field does not exist, fill with defaults or raise error.
3.- Use that structure to create actual `desiredSecret`. 
4.- Reconcile secret

pkg/controller/apicast/apicast_logic_reconciler.go
miguelsorianod: nope. It was moved to the controller itself. I'll remove it :+1:

miguelsorianod:
I've updated the PR to join the accesstoken and url field in the same secret field and only use one secret for AdminPortal.

I've also done some refactorings

pkg/controller/apicast/apicast_logic_reconciler.go
```diff
@@ -0,0 +1,423 @@
...
+	}
+	_, ok = parsedURL.User.Password()
+	if !ok {
+		return nil, fmt.Errorf("Access Token required in %s URL", apicast.AdminPortalURLAttributeName)
+	}
+
+	return &adminPortalCredentialsSecret, err
+}
+
+func (r *APIcastLogicReconciler) reconcileAdminPortalCredentials() (*v1.Secret, bool, error) {
```
miguelsorianod: This method also returns a bool that is the "changed" flag.
In case OwnerRef is updated we requeue and not continue in this specific case of an "external" user-provided secret

pkg/apicast/apicast.go
eguzki: Yes, It allows you to convert, filter process the CR information to something else. Besides you can get information not only from CR, but also from ConfigMaps and Secrets. Set them in the internal struct and then use it to build desired k8s objects.

pkg/controller/apicast/apicast_logic_reconciler.go
```diff
@@ -0,0 +1,423 @@
...
+		ResponseCodesIncluded:            r.APIcastCR.Spec.ResponseCodesIncluded,
+		CacheConfigurationSeconds:        r.APIcastCR.Spec.CacheConfigurationSeconds,
+		ManagementAPIScope:               managementAPIScope,
+		OpenSSLPeerVerificationEnabled:   r.APIcastCR.Spec.OpenSSLPeerVerificationEnabled,
+	}
+
+	return apicastResult, err
+}
+
+func (r *APIcastLogicReconciler) namespacedName(object metav1.Object) types.NamespacedName {
```
eguzki: This method does not make sense as `APIcastLogicReconciler` method

pkg/controller/apicast/apicast_logic_reconciler.go
eguzki: I think you should. Maybe not in this PR. 

Or, reconcile one by one. Each approach has trade-offs.

pkg/apicast/apicast.go
```diff
@@ -0,0 +1,308 @@
...
+	Namespace                        string
+	DeploymentName                   string
+	ServiceName                      string
+	Replicas                         int32
+	AppLabel                         string
+	AdditionalAnnotations            map[string]string
+	ServiceAccountName               string
+	Image                            string
+	ExposedHost                      ExposedHost
+	OwnerReference                   *metav1.OwnerReference
```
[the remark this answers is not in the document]
miguelsorianod: I made it this way so we could create an APIcast without having an OwnerReference.
That's why it's a pointer to a metav1.OwnerReference

pkg/controller/apicast/apicast_logic_reconciler.go
miguelsorianod: I put it there because currently it is used there. In case in the future we need it in another place we can refactor it if needed and think a new location for it.

miguelsorianod:
I've updated the PR to use LocalObjectReference instead of SecretReference for the AdminPortalCredentialsRef attribute in apicast_types.go

The reason for this is that when an operator is namespaced scope (watch changes in a specific namespace) it only watches changes of the object in the specified namespace. The client provided by the operator-sdk also is only able to retrieve objects from the specified namespace and not from other namespaces. So, it only makes sense to provide a name for the secret, as we wouldn't be able to retrieve/watch changes of a secret in another namespace.

miguelsorianod:
Regarding sorting the arrays, I'll do it in another PR :+1:

pkg/controller/apicast/apicast_logic_reconciler.go
eguzki: This kind of utility methods should always be defined externally.

The bad smell is very clear when you do not use the method receiver.

It's ok to leave there for now

PR #2, opened 2019-09-05
Apicast add gateway embedded config

miguelsorianod:
This PR allows the configurability of an embedded APIcast configuration through the operator

### DIFF
--- a/pkg/apis/apps/v1alpha1/apicast_types.go
+++ b/pkg/apis/apps/v1alpha1/apicast_types.go
@@ -16,15 +16,17 @@ type APIcastSpec struct {
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
 	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
 	// +optional
-	Replicas                  *int64                   `json:"replicas,omitempty"`
-	AdminPortalCredentialsRef *v1.LocalObjectReference `json:"adminPortalCredentialsRef"`
+	Replicas *int64 `json:"replicas,omitempty"`
+	// +optional
+	AdminPortalCredentialsRef *v1.LocalObjectReference `json:"adminPortalCredentialsRef,omitempty"`
+	// +optional
+	EmbeddedConfigurationSecretRef *v1.LocalObjectReference `json:"embeddedConfigurationSecretRef,omitempty"`
 	// +optional
 	ServiceAccount *string `json:"serviceAccount,omitempty"`
 	// +optional
 	Image *string `json:"image,omitempty"`
 	// +optional
 	ExposedHost *APIcastExposedHost `json:"exposedHost,omitempty"`
-
 	// +optional
 	DeploymentEnvironment *DeploymentEnvironmentType `json:"deploymentEnvironment,omitempty"` // THREESCALE_DEPLOYMENT_ENV
 	// +optional
--- a/pkg/controller/apicast/apicast_logic_reconciler.go
+++ b/pkg/controller/apicast/apicast_logic_reconciler.go
@@ -26,7 +26,8 @@ import (
 )
 
 const (
-	AdmPortalSecretResverAnnotation = "apicast.apps.3scale.net/admin-portal-secret-resource-version"
+	AdmPortalSecretResverAnnotation            = "apicast.apps.3scale.net/admin-portal-secret-resource-version"
+	GatewayConfigurationSecretResverAnnotation = "apicast.apps.3scale.net/gateway-configuration-secret-resource-version"
 )
 
 type APIcastLogicReconciler struct {
@@ -36,6 +37,7 @@ type APIcastLogicReconciler struct {
 
 type apicastUserProvidedSecrets struct {
 	adminPortalCredentialsSecret *v1.Secret
+	gatewayEmbeddedConfigSecret  *v1.Secret
 }
 
 func NewAPIcastLogicReconciler(b BaseReconciler, cr *appsv1alpha1.APIcast) APIcastLogicReconciler {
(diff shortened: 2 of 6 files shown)


pkg/apis/apps/v1alpha1/apicast_types.go
```diff
@@ -16,15 +16,17 @@ type APIcastSpec struct {
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
 	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
 	// +optional
-	Replicas                  *int64                   `json:"replicas,omitempty"`
-	AdminPortalCredentialsRef *v1.LocalObjectReference `json:"adminPortalCredentialsRef"`
+	Replicas *int64 `json:"replicas,omitempty"`
+	// +optional
+	AdminPortalCredentialsRef *v1.LocalObjectReference `json:"adminPortalCredentialsRef,omitempty"`
+	// +optional
+	GatewayConfigurationSecret *v1.LocalObjectReference `json:"gatewayConfigurationSecret,omitempty"`
```
miguelsorianod: I am not sure what would be a good name for it.
Do you think is there a better name to design "the secret that the user must provide to specify an apicast configuration (json config of apicast) ?

pkg/apis/apps/v1alpha1/apicast_types.go
```diff
@@ -16,15 +16,17 @@ type APIcastSpec struct {
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
 	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
 	// +optional
-	Replicas                  *int64                   `json:"replicas,omitempty"`
-	AdminPortalCredentialsRef *v1.LocalObjectReference `json:"adminPortalCredentialsRef"`
+	Replicas *int64 `json:"replicas,omitempty"`
+	// +optional
+	AdminPortalCredentialsRef *v1.LocalObjectReference `json:"adminPortalCredentialsRef,omitempty"`
```
miguelsorianod: I have changed AdminPortalCredentialsRef to be optional instead of mandatory, because now you can deploy apicast specifying the AdminPortal credentials or the Apicast configuration in json in itself (via another secret).

However, this has a downside:
By making both elements optional the user might wrongly deploy an Apicast without setting any of them.
I've added code so the operator returns an error when this happens but it's not ideal.
What do you think about this?

pkg/controller/apicast/apicast_logic_reconciler.go
```diff
@@ -63,6 +65,10 @@ func (r *APIcastLogicReconciler) Reconcile() (reconcile.Result, error) {
 		return reconcile.Result{}, nil
 	}
 
+	if r.APIcastCR.Spec.AdminPortalCredentialsRef == nil && r.APIcastCR.Spec.GatewayConfigurationSecret == nil {
```
miguelsorianod: Here I check that one of the secrets must be set

pkg/controller/apicast/apicast_logic_reconciler.go
```diff
@@ -140,7 +155,52 @@ func (r *APIcastLogicReconciler) getAdminPortalCredentialsSecret() (*v1.Secret,
...
+		return v1.Secret{}, err
+	}
+	if _, ok := gatewayConfigSecret.Data["config.json"]; !ok {
+		return v1.Secret{}, fmt.Errorf("%s key not found in secret %s", "config.json", gatewayConfigSecret.Name)
+	}
+	return gatewayConfigSecret, nil
+}
+
 func (r *APIcastLogicReconciler) reconcileAdminPortalCredentials() (*v1.Secret, bool, error) {
+	if r.APIcastCR.Spec.AdminPortalCredentialsRef == nil {
```
miguelsorianod: If the secret has not been specified I return at this point.
We could make the if outside of this method and do it in the "Reconcile" method not sure which of the two options would be better. If we were to do it in reconcile then we would have to put an if there and perform the corresponding checks which I didn't like that much

miguelsorianod:
Ready for review @eguzki

pkg/apis/apps/v1alpha1/apicast_types.go
eguzki: and let apicast do that?

pkg/apis/apps/v1alpha1/apicast_types.go
eguzki: Env var is called  `THREESCALE_CONFIG_FILE`

So maybe `ThreeScaleConfigSecret` ???

pkg/controller/apicast/apicast_logic_reconciler.go
eguzki: see above

pkg/controller/apicast/apicast_logic_reconciler.go
```diff
@@ -140,7 +155,52 @@ func (r *APIcastLogicReconciler) getAdminPortalCredentialsSecret() (*v1.Secret,
 	return &adminPortalCredentialsSecret, err
 }
 
+func (r *APIcastLogicReconciler) reconcileGatewayConfigurationSecret() (v1.Secret, error) {
```
eguzki: Remove unused methods

pkg/controller/apicast/apicast_logic_reconciler.go
eguzki: IMO it's ok calling this method and return early

pkg/controller/apicast/apicast_logic_reconciler.go
miguelsorianod: :+1:

pkg/apis/apps/v1alpha1/apicast_types.go
miguelsorianod: I don't like to have the name ThreeScale where we are already in the context of a 3scale related CR. I understand you have said this name because the envvar is the same but I prefer to making more meaningful / context-related names. The user does not care about envvars but about the effects that they do.

I'll change the name to EmbeddedConfigurationSecret.

miguelsorianod:
Ready to review

pkg/apis/apps/v1alpha1/apicast_types.go
miguelsorianod: As we talked, for the moment we'll check it in the operator so it "breaks" before.
We'll see whether we could control this with OpenAPI in the future. It might require restructuration of the fields though

miguelsorianod: Thinking more about it I removed the check in the operator, so we do not tie this behaviour into it in case we want to change it in the future with OpenAPI.

PR #4, opened 2019-09-09
Add Apache 2 LICENSE and NOTICE files

miguelsorianod:


eguzki:
Would be nice to add `licenses.xml` with depencency licenses

miguelsorianod:
:+1: 

The targets are already there but I'll generate the licenses.xml. I'll do it in another PR to differentiate those changes from the Apache2 license addition

PR #9, opened 2019-09-09
Apicast operator doc

eguzki:


deploy/crds/apps_v1alpha1_apicast_crd.yaml
```diff
@@ -80,6 +80,15 @@ spec:
             serviceAccount:
               type: string
           type: object
+          anyOf:
```
miguelsorianod: This AnyOf is applied in the "Spec" section for all the fields in them.

Doing this makes not possible to add validation for other set of fields.
For example if we were to have two other fields where we wanted to make sure one of the two (or the two) had to be set and we add it in this anyOf then it would have 4 elements.
That's not what we intend.

In any case we should move the adminPortal and the embeddedConfig fields into a new section of the spec (for example configurationSource) and there define the two fields and add the anyOf in that section.

Also we should talk whether we want to do this. There are some situations that might happen in the future that would break this validation. For example if you have three fields A,B,C and A and B are compatible but A and C are not. In that scenario AnyOf would not work because the combination A,C is not compatible (this might be an edge case but can theoretically happen)

README.md
```diff
@@ -34,16 +34,130 @@ that the operator is able to understand and process.
 * [git][git_tool]
 * [go] version 1.12.5+
 * [kubernetes] version v1.11.0+
-* [oc] version v3.11+
+* [kubectl] version v1.11+
```
miguelsorianod: I'd directly put 1.13+ as minimum version.
In that way we make sure we follow current OpenShift release (4.1) and we are not tied to having to support older versions.
It's interesting to have 1.13+ as 1.11 did not support schema changes, did not have webhook conversions etc.

README.md
```diff
@@ -34,16 +34,130 @@ that the operator is able to understand and process.
...
+Create [apicast custom resource](doc/apicast-crd-reference.md):
+
+```yaml
+$ cat my-echo-apicast.yaml
+apiVersion: apps.3scale.net/v1alpha1
+kind: APIcast
+metadata:
+  name: my-echo-apicast
+spec:
+  exposedHost:
```
miguelsorianod: Maybe we shouldn't specify exposedHost in the basic example?
It's not really necessary (only if you want it to be externally accessible) and depending on the situation (if they don't have a default ingress controller set) they should create an Ingress Controller so it might not be ideal for a "basic" scenario

README.md
miguelsorianod: Same applies for kubernetes version

README.md
```diff
@@ -34,16 +34,130 @@ that the operator is able to understand and process.
 * [git][git_tool]
 * [go] version 1.12.5+
 * [kubernetes] version v1.11.0+
-* [oc] version v3.11+
+* [kubectl] version v1.11+
 * Access to a Openshift v3.11.0+ cluster.
```
miguelsorianod: That would change. To a Kubernetes cluster

README.md
```diff
@@ -34,16 +34,130 @@ that the operator is able to understand and process.
...
 * A user with administrative privileges in the OpenShift cluster.
```
miguelsorianod: in the Kubernetes cluster

deploy/crds/apps_v1alpha1_apicast_crd.yaml
eguzki: That is not true.

`anyOf` is a conditional to apply subschemas. It mean, any of the following must happen, otherwise validation fails. it does NOT check ALL the attributes.

Besides, there is no need (unless there is another reason to do) to group in a new section, structural schemas (based on openapi 3 validation) is flexible enough. For instance, you may have another conditional subschema to say, for instance, `anyOf` `image` or `logLevel` should be present like this:

```
anyOf:                                            
  - properties:                                   
      adminPortalCredentialsRef:                  
        type: object                              
    required: ["adminPortalCredentialsRef"]       
  - properties:                                   
        type: object                              
    required: ["embeddedConfigurationSecretRef"]  
anyOf:                                            
  - properties:                                   
      image:                                      
        type: string                              
    required: ["image"]                           
  - properties:                                   
      logLevel:                                   
        type: string                              
    required: ["logLevel"]                        
```

Anyway, having this enforcement at CRD level (cluster level) and not at operator level is something worth to be discussed in other PR. I will open another PR enforcing any of the following configuration sources to exist, `adminPortalCredentialsRef` or `embeddedConfigurationSecretRef`

eguzki: https://github.com/3scale/apicast-operator/pull/10

README.md
eguzki: :+1:

eguzki: 1.11 does support schema changes, but in a limited way. Once you update your CRD, older apiversions have to be update as well.

miguelsorianod: Even if you "can do that" I don't think you want to have to change the original schema if you increase the APIversion. What's the point otherwise? In that case you would be changing the original version and that's it.
Also, no ability to perform incompatible changes, which is one of the other reasons to increase APIVersion

deploy/crds/apps_v1alpha1_apicast_crd.yaml
miguelsorianod: As far as I know It DOES check all attributes. Between all attributes in spec it checks that any of the specified ones in AnyOf are specified in spec.

Regarding specifying multiple times anyOf :+1: if you have seen we could do this then if we decide to do that would be a good option :+1: . Have you found this behavior described in the documentation somewhere?

README.md
eguzki: It means you cannot have two apiversions active at a given point of time. But you can upgrade.

Breaking compatibility of previous  CRD versions has shown to be quite an issue. You need to provide schema conversion webhook and make sure no other API is depending on your API. You need a very good reason to do so, even in 1.13

README.md
eguzki: For me it does not make sense deploying apicast without ingress resource, i.e. not being exposed externally. Does it? This makes me think that maybe CRD should **require** exposedHost (I would call it `domainName`). WDYT?

Which ingress controller they use does not matter... this is another topic. Am I right?

README.md
miguelsorianod: > It means you cannot have two apiversions active at a given point of time. But you can upgrade.

What do you mean with this? You can be serving several APIversions at the same time in Kubernetes for the same CRD. I'm not sure I'm understanding what you say

> Breaking compatibility of previous CRD versions has shown to be quite an issue. You need to provide schema conversion webhook and make sure no other API is depending on your API. You need a very good reason to do so, even in 1.13

That is not an argument against requiring 1.13+, just at the contrary, in case we do that (we should not do it lightly but we should be able to do it  ) we need 1.13+ : )

README.md
miguelsorianod: To me it makes sense.

You could be using apicast and connecting it with other elements in the Kubernetes cluster.

I don't think we should require exposedHost at all

deploy/crds/apps_v1alpha1_apicast_crd.yaml
eguzki: Conditions check for the specified fields on the current context. It does NOT check all attributes. That is why conditions expect a list of properties.

From [Openapi Docs](https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/). The hint that made me try was the text in the doc `validate a value against multiple criteria.`

miguelsorianod: But you have to check those conditions from a set of fields. In this case, it is checked that among all the fields in the spec section any of them (1 or more) are the ones specified in the anyOf section

README.md
eguzki: `All specified versions must use the same schema`

miguelsorianod: That does not mean you cannot have two apiversions active at the same time. It says that all apiversions should have the same schema

eguzki: ok, fair enough, you are right. to be specific you cannot have two schemas active but you can have to apiversion active :man_facepalming: 

but YOU CAN UPGRADE not breaking comp.

So, supporting 1.11 should not be discarded. Let's product decide, for now we keep open to all possibilities

miguelsorianod: We can discuss it with product of course, but to me there are clearly technical specific characteristics that should be totally our responsability and that we should be responsible for them.

I think an operator should be able to perform incompatible changes at schema level and it should be able to have APIversions with different schemas. Kubernetes itself has that in their native objects, and so do several other operators. We have also found evidence that we might want to do that with 3scale-operator and we should be able to adapt to possible new changes in this case too that require that.

README.md
eguzki: It seems we cannot predict what the use cases are going to be. We will leave `exposedHost`  as optional. Still valid the proposal of changing name to `domainName`.

However, I would leave the example as it is because it is a "getting started" basic example and you probably want to curl it from your terminal, hence needs to be exposed.

miguelsorianod: You don't really need to expose to curl from the terminal. You can proxy the pod / service through kubectl port-forward. It might be easier to just expose the host so we can leave it that way if you like it, although the user will have to have to set a resolvable DNS name in that case.

The reason I chose exposedHost (and host as the subfield) and not domain is due to I identify "domain" with the name given to the "network domain" and host to the specific host in that domain.

eguzki:
REady to be merged @miguelsorianod

README.md
```diff
@@ -34,16 +29,130 @@ that the operator is able to understand and process.
...
+mkdir -p $GOPATH/src/github.com/3scale
+cd $GOPATH/src/github.com/3scale
+git clone https://github.com/3scale/apicast-operator.git
+cd apicast-operator
+```
+
+With administrative user create and deploy a Apicast CRD
+
+```sh
+kubectl delete -f deploy/crds/apps_v1alpha1_apicast_crd.yaml
```
miguelsorianod: this should be create

eguzki: :+1:

eguzki:
ready  @miguelsorianod

miguelsorianod:
As we commented, maybe "Domain" is not the appropriate name to explain the "host" field in the documentation, although the "exposedHost" might not be a good name either.
In any case, lets handle that in another PR/issue. This PR looks good to me, but lets update the documentation if necessary later.

PR #11, opened 2019-09-10
fix configuration loader processing

eguzki:


pkg/apis/apps/v1alpha1/apicast_types.go
```diff
@@ -34,7 +34,8 @@ type APIcastSpec struct {
 	// +optional
 	EnabledServices []string `json:"enabledServices,omitempty"` // APICAST_SERVICES_LIST
 	// +optional
-	ConfigurationLoadMode *int64 `json:"configurationLoadMode,omitempty"` // APICAST_CONFIGURATION_LOADER
+	// +kubebuilder:validation:Enum=boot;lazy
```
miguelsorianod: I see you relied in the kubebuilder comment tag as we talked. I'm not sure whether the behaviour for it will change in the future in operator-sdk but lets give it a try.

PR #12, opened 2019-09-10
fix log level param parsing

eguzki:


pkg/apis/apps/v1alpha1/apicast_types.go
```diff
@@ -36,7 +36,8 @@ type APIcastSpec struct {
 	// +optional
 	ConfigurationLoadMode *int64 `json:"configurationLoadMode,omitempty"` // APICAST_CONFIGURATION_LOADER
 	// +optional
-	LogLevel *LogLevelType `json:"logLevel,omitempty"` // APICAST_LOG_LEVEL
+	// +kubebuilder:validation:Enum=debug;info;notice;warn;error;crit;alert;emerg
+	LogLevel *string `json:"logLevel,omitempty"` // APICAST_LOG_LEVEL
```
miguelsorianod: If we decide to replace the typed string for a pure string then the following part of code in apicast_logic_reconciler.go is not needed anymore

```
	var logLevel *string
	if r.APIcastCR.Spec.LogLevel != nil {
		res := string(*r.APIcastCR.Spec.LogLevel)
		logLevel = &res
	}
```

eguzki:

eguzki:
"close and comment" button should have confirmation XD

PR #13, opened 2019-09-10
fix management api param processing

eguzki:


pkg/apis/apps/v1alpha1/apicast_types.go
```diff
@@ -44,7 +44,8 @@ type APIcastSpec struct {
 	// +optional
 	CacheConfigurationSeconds *int64 `json:"cacheConfigurationSeconds,omitempty"` // APICAST_CONFIGURATION_CACHE
 	// +optional
-	ManagementAPIScope *ManagementAPIScopeType `json:"managementAPIScope,omitempty"` // MANAGEMENT_API
+	// +kubebuilder:validation:Enum=disabled;status;policies;debug
+	ManagementAPIScope *string `json:"managementAPIScope,omitempty"` // MANAGEMENT_API
```
miguelsorianod: Nitpic. Can we fix the comment here too?

miguelsorianod: If we change the typed string into string then the follwing content in apicast_logic_reconciler.go is not needed anymore:

```

	var managementAPIScope *string
	if r.APIcastCR.Spec.ManagementAPIScope != nil {
		res := string(*r.APIcastCR.Spec.ManagementAPIScope)
		managementAPIScope = &res
	}
```

PR #14, opened 2019-09-10
Enhance apicast reconciliation

eguzki:
Reconcile Labels
Reconcile Deployment environment array with set approach

### DIFF
--- a/pkg/controller/apicast/apicast_logic_reconciler.go
+++ b/pkg/controller/apicast/apicast_logic_reconciler.go
@@ -189,6 +189,7 @@ func (r *APIcastLogicReconciler) reconcileAdminPortalCredentials() (*v1.Secret, 
 	}
 
 	if changed {
+		r.Logger().Info(fmt.Sprintf("Updating %s", k8sutils.ObjectInfo(adminPortalCredentialsSecret)))
 		err = r.Client().Update(context.TODO(), adminPortalCredentialsSecret)
 		if err != nil {
 			return nil, changed, err
@@ -214,6 +215,7 @@ func (r *APIcastLogicReconciler) reconcileGatewayEmbbededConfig() (*v1.Secret, b
 	}
 
 	if changed {
+		r.Logger().Info(fmt.Sprintf("Updating %s", k8sutils.ObjectInfo(gatewayEmbeddedConfigSecret)))
 		err = r.Client().Update(context.TODO(), gatewayEmbeddedConfigSecret)
 		if err != nil {
 			return nil, changed, err
@@ -350,6 +352,7 @@ func (r *APIcastLogicReconciler) namespacedName(object metav1.Object) types.Name
 
 func (r *APIcastLogicReconciler) initialize() (bool, error) {
 	if appliedSomeInitialization := r.applyInitialization(); appliedSomeInitialization {
+		r.Logger().Info(fmt.Sprintf("Updating %s", k8sutils.ObjectInfo(r.APIcastCR)))
 		err := r.Client().Update(context.TODO(), r.APIcastCR)
 		if err != nil {
 			return false, err
@@ -418,8 +421,8 @@ func (r *APIcastLogicReconciler) reconcileDeployment(desiredDeployment appsv1.De
 	err := r.Client().Get(context.TODO(), r.namespacedName(&desiredDeployment), &existingDeployment)
 	if err != nil {
 		if errors.IsNotFound(err) {
+			r.Logger().Info(fmt.Sprintf("Creating %s", k8sutils.ObjectInfo(&desiredDeployment)))
 			err = r.Client().Create(context.TODO(), &desiredDeployment)
-			r.Logger().Info("Creating Deployment...")
 			return err
 		}
 		return err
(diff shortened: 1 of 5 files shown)


pkg/controller/apicast/apicast_logic_reconciler.go
```diff
@@ -418,40 +421,34 @@ func (r *APIcastLogicReconciler) reconcileDeployment(desiredDeployment appsv1.De
...
+	updatedTmp := ReconcileEnvVar(&existingDeployment.Spec.Template.Spec.Containers[0].Env, desiredDeployment.Spec.Template.Spec.Containers[0].Env)
+	changed = changed || updatedTmp
 
-	// TODO should we merge instead of set the desired annotations? Keep in mind that this is are annotations
-	// of the PodTemplate, not the Pod itself. Can a PodTemplate be modified by other elements?
-	if !reflect.DeepEqual(existingDeployment.Spec.Template.Annotations, desiredDeployment.Spec.Template.Annotations) {
-		changed = true
-		existingDeployment.Spec.Template.Annotations = desiredDeployment.Spec.Template.Annotations
-	}
+	updatedTmp = k8sutils.EnsureObjectMeta(&existingDeployment.Spec.Template.ObjectMeta, &desiredDeployment.Spec.Template.ObjectMeta)
```
miguelsorianod: This has a big impact on the effect of the reconciliation in this case.

Until now, when an annotation disappeared from the desired then the existing was updated accordingly and the annotation removed. For example, the "apicast.apps.3scale.net/admin-portal-secret-resource-version" annotation is set when AdminPortal is defined in the CR.
Before this PR, if you had the AdminPortal defined in the CR and then you unset the AdminPortal from the CR, the annotation (with its value) was also removed from the PodTemplate spec. With the new behavior (merge) the annotation will be left in the PodTemplate spec instead of removed.
Are we ok with that?

Also, be aware as I tried to explain in the TODO code comment that this specific case might not be the same as other annotation/labels ObjectMeta. This ObjectMeta is inside a PodTemplate section. This is used as the base (template) when creating pods. I don't know whether other controllers might modify this sections (this can happen in "normal" annotations/labels sections because they are applied on existing objects, this ObjectMeta is just for the PodTemplate spec)

It seems that adding merge behavior to any field/section means that when removing functionality from the desired it remains there unless you start to implement explicit logic to remove it before applying the merge. For example, this behavior would happen if you try to merge-reconcile an array of volumes too. If you were to have a flag that controls if  a volume exists or not (indirectly) through the CR, as we do with the embeddedConfiguration flag and we merge-reconcile the volumes, then if the user decides to remove that embeddedConfiguration field the volume would be left there even not being used.

pkg/controller/apicast/apicast_logic_reconciler.go
```diff
@@ -418,40 +421,34 @@ func (r *APIcastLogicReconciler) reconcileDeployment(desiredDeployment appsv1.De
...
+	if existingDeployment.Spec.Template.Spec.ServiceAccountName != desiredDeployment.Spec.Template.Spec.ServiceAccountName {
 		changed = true
 		existingDeployment.Spec.Template.Spec.ServiceAccountName = desiredDeployment.Spec.Template.Spec.ServiceAccountName
 	}
 
-	if !reflect.DeepEqual(existingDeployment.Spec.Template.Spec.Containers[0].Env, desiredDeployment.Spec.Template.Spec.Containers[0].Env) {
-		changed = true
-		existingDeployment.Spec.Template.Spec.Containers[0].Env = desiredDeployment.Spec.Template.Spec.Containers[0].Env
-	}
+	updatedTmp := ReconcileEnvVar(&existingDeployment.Spec.Template.Spec.Containers[0].Env, desiredDeployment.Spec.Template.Spec.Containers[0].Env)
```
miguelsorianod: The same concern/doubt appears here as the comment I've left below.
This has even bigger impact as this ends up as environment variables directly in the deployment.

Lets say you were to change from EmbeddedConfiguration config source to AdminPortal config source in the CR.
If I'm not wrong this would leave the EmbeddedConfiguration corresponding environment variable in the DeploymentConfig and also add the new one for AdminPortal. As EmbeddedConfiguration envvar has precedence over the AdminPortal one the apicast deployment would not behave as desired and would it be a bug?

eguzki: totally right, I am going to refactor this code

pkg/controller/apicast/apicast_logic_reconciler.go
eguzki: Indeed, merge approach does not work when you remove something from source resource and you want it removed from target resource as well. It is well known. I think the approach fits well for the annotations and labels case. I also think it does not fit well for the array of environment variables as I answered to your comment above.  I see you have some doubts. I do not see any reason  we should not be using merge approach for annotations and labels. But let me know and I can change reconciliation approach.

`apicast.apps.3scale.net/admin-portal-secret-resource-version` and `apicast.apps.3scale.net/gateway-configuration-secret-resource-version` annotations are not being used, but set in annotations. Do you think we should fix that and do no set it?

Merge approach is not being used when reconciling volumes, mainly because you need to remove them when unused.

eguzki: It is not expected any controller to update them, so we use "set" approach, instead of merge. This way any removed annotation from desired (due to change in CR) will be removed in existing too.

eguzki:

pkg/controller/apicast/apicast_logic_reconciler.go
```diff
@@ -418,41 +421,45 @@ func (r *APIcastLogicReconciler) reconcileDeployment(desiredDeployment appsv1.De
...
-	// of the PodTemplate, not the Pod itself. Can a PodTemplate be modified by other elements?
+	// They are annotations of the PodTemplate, part of the Spec, not part of the meta info of the Pod or Environment object itself
+	// It is not expected any controller to update them, so we use "set" approach, instead of merge.
+	// This way any removed annotation from desired (due to change in CR) will be removed in existing too.
 	if !reflect.DeepEqual(existingDeployment.Spec.Template.Annotations, desiredDeployment.Spec.Template.Annotations) {
 		changed = true
 		existingDeployment.Spec.Template.Annotations = desiredDeployment.Spec.Template.Annotations
 	}
 
+	if !reflect.DeepEqual(existingDeployment.Spec.Template.Labels, desiredDeployment.Spec.Template.Labels) {
```
miguelsorianod: Why are we reconciling Deployment's PodTemplate labels?

If I'm not wrong, we are not providing configuration of them through the CR.

eguzki: `apicastFullName := "apicast-" + r.APIcastCR.Name`

eguzki: I guess we do

miguelsorianod: You cannot change a CR name once deployed so no point in doing so

eguzki: why not?? then why it is exposed?

miguelsorianod: It's not exposed. All object types have a name

miguelsorianod: You cannot change a Kubernetes object name once it is deployed.

eguzki: right, no label reconciliation then

eguzki:

PR #23, opened 2019-09-30
remove ownership to resources not created by operator

eguzki:
Remove ownership to resources not created by operator.

They can be reused by two different CR. For instance, production and staging apicast resources.

### DIFF
--- a/pkg/controller/apicast/apicast_logic_reconciler.go
+++ b/pkg/controller/apicast/apicast_logic_reconciler.go
@@ -65,20 +65,25 @@ func (r *APIcastLogicReconciler) Reconcile() (reconcile.Result, error) {
 		return reconcile.Result{}, nil
 	}
 
-	adminPortalCredentialsSecret, changed, err := r.reconcileAdminPortalCredentials()
-	if err != nil {
-		return reconcile.Result{}, err
-	}
-	if changed {
-		return reconcile.Result{Requeue: true}, nil
+	// CRD validations ensures either
+	// r.APIcastCR.Spec.AdminPortalCredentialsRef
+	// or
+	// r.APIcastCR.Spec.EmbeddedConfigurationSecretRef
+	// are not nil
+	var adminPortalCredentialsSecret *v1.Secret
+	if r.APIcastCR.Spec.AdminPortalCredentialsRef != nil {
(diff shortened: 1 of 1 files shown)


miguelsorianod:
The reason we add OwnerReference to those secrets is to be able to react to changes to them via the OwnerReference when watching them.
Removing them as done here would remove the ability to perform reconciliation when those change and we don't want to lose that.

eguzki:
True, this PR in incomplete. 

Thinking about this, the following discussion comes up. There are two feasible options:
* Option A: Currently implemented option. The secret cannot be shared between two different apicast resources (multiple ownership is not allowed by k8s). The apicast operator reconciliates the information *inside* the secret. Thus, if the information of that secret changes, operator will redeploy apicast deployment updating some annotation with latest resource version of the secret resource.
* Option B: Externally created secrets are not owned by a single CR, thus can be shared between CR's. Apicast operator does NOT reconciliate the information contained by the secret. Instead, the operator reconciliates secret itself. So, to get same functionality, if user wants to update 3scale admin portal URL, the user needs to create a new secret with the updated information. In order to let the operator know about the change (and have it reconciliated), update CR's field value pointing to the secret, like the example below.
```
spec:
  adminPortalCredentialsRef:
     name: myapicast-secret -> Update to my-new-secret
```

Those are two feasible options. 

I am more inclined to option B. Feels like more intuitive and flexible. Not sharing secrets or configmaps seems like a bad smell. In addition, option B removes the procedure of updating annotations of the deployment to redeploy the apicast, because the rest does not change. With option B, deployment spec would change (having Env Var coming from a *new* secret), thus redeployment. It requires, though, to be clearly documented that to update portal URL, the user needs to edit CR, pointing to another secret (recently created or an old one)

eguzki:
Actually, `Object` interface allows an array of owner references. Potentially this mean you could set multiple owner references to one resource. It is not clear if k8s allows multiple owners from [owners and dependents doc](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#owners-and-dependents)

However, [controller runtime libra enforces](https://github.com/kubernetes-sigs/controller-runtime/blob/801e12a501606245062a6ff77ac91eef8936d9fb/pkg/controller/controllerutil/controllerutil.go#L76) enforces single ownership. So, independently of what k8s allows, currently we cannot set multiple ownership.

PR #28, opened 2019-10-02
E2e test on k8s

eguzki:


.circleci/config.yml
```diff
@@ -204,7 +252,35 @@ jobs:
...
+            docker load -i /tmp/workspace/images/operator-image.tar
+      - kubernetes/create-or-update-resource:
+          resource-file-path: test/resources/namespace-operator-test.yaml
+          show-kubectl-command: true
+      - watch-operator-pod
+      - watch-operator-pod-logs
+      - run:
+          name: run E2E tests
+          command: |
+            make e2e-run IMAGE=172.30.1.1:5000/openshift/apicast-operator:${CIRCLE_SHA1} NAMESPACE=operator-test
```
miguelsorianod: In other places of the file I've seen references to 172.30.1.1:5000/openshift/apicast-operator:test instead of ${CIRCLE_SHA1}. Specifically in  tag-operator-image-master and tag-operator-image-release jobs. Shouldn't be all :test or all :{CIRCLE_SHA1} ?

deploy/operator.yaml
```diff
@@ -19,7 +19,6 @@ spec:
           image: REPLACE_IMAGE
           command:
           - apicast-operator
-          imagePullPolicy: Always
```
miguelsorianod: When you create a project with the operator-sdk it automatically creates the deploy/operator.yaml with this configuration.
Is there a reason we want to remove this? Is it because you don't want the image to be always pulled?
We have the following tradeoffs:
 * If imagePullPolicy is not set then it will not always download the image (faster)
 * If imagePullPolicy is not set then if the  image tag is refreshed we won't get the new changes (will miss changes).

.circleci/config.yml
eguzki: :+1:

deploy/operator.yaml
eguzki: three reasons:
* Apicast operator image will not be a floating tag, but fixed tag, so we know what is exact image running and we can manage upgrades.
* Testing "pushes" image to internal registry, so we need to be able to fetch from internal registry.
* As you know, default imagepullpolicy (IfNotPresent) will always download image if tag is latest or no tag is set. So we can mimic behavior if required. I think default policy is the most flexible and efficient one when using fixed tags.

eguzki:

PR #39, opened 2019-12-13
Fix settings defaults for image and accountservice

eguzki:
It does not make sense to set defaults on some fields at the custom resource. If the operator sets some default in the custom resource, on future reconciliation runs there is no way to know if the value has been set by the user or not. This procedure can be useful for some fields, as the user can see what defaults are being used. But, on the other hand, it can entirely disable upgrading mechanism for some cases, for instance, when that field is the image url of the product.

The operator cannot change the desired state and if image is part of the desired state, upgrading cannot be performed. And the operator does not have any way of knowing if the value has been set as default value or has been set by the user.

Setting defaults has other drawback, as the operator cannot change them (because they are wrong or outdated or whatever) when operator is upgraded. Once the defaults are written to the CR, those values are no longer default values, but desired values.

ServiceAccount name's default value was removed from the CR for that reason.

### DIFF
--- a/pkg/controller/apicast/apicast_logic_reconciler.go
+++ b/pkg/controller/apicast/apicast_logic_reconciler.go
@@ -317,14 +317,24 @@ func (r *APIcastLogicReconciler) internalAPIcast(userProvidedSecrets *apicastUse
 		gatewayConfigurationSecretName = &tmpGatewayConfigurationSecretName
 	}
 
+	image := apicast.GetDefaultImageVersion()
+	if r.APIcastCR.Spec.Image != nil {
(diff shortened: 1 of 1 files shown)


miguelsorianod:
Ok to the changes.
Only point is that it hides the value of some useful information for the user/debug purposes that can be configured at the spec level. I guess the solution is add it into the "status" field in that case. Should do we add it at least for the image field? what do you think? In any case independently of the response I think that can be added in another PR if we decide to do so :+1:

eguzki:
Yup, the current image value in the status field would be very nice.

PR #41, opened 2019-12-19
Add image field to APIcast status

miguelsorianod:


pkg/apis/apps/v1alpha1/apicast_types.go
```diff
@@ -71,6 +71,10 @@ type APIcastStatus struct {
 	// +patchMergeKey=type
 	// +patchStrategy=merge
 	Conditions []APIcastCondition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
+
+	// The image being used in the APIcast deployment
+	// +optional
+	Image string `json:"image,omitempty"`
```
miguelsorianod: I've set this field as optional and omitempty.
The reason for that is that I've seen other Kubernetes objects that also havfe some status fields that seem they should always be defined but they also are optional,omitempty. For example, The replicas field in a Deployment is optional,omitempty

pkg/controller/apicast/apicast_controller.go
```diff
@@ -189,6 +191,41 @@ func (r *ReconcileAPIcast) Reconcile(request reconcile.Request) (reconcile.Resul
...
+	// Check if the image field in the Deployment object matches the field in the CR.
+	//   If it does match we do nothing
+	//   If it does not match we update the CR and requeue
+	apicast, err := reconciler.APIcastFromCRContents()
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	var apicastDeployment *appsv1.Deployment
+	r.Client().Get(context.TODO(), types.NamespacedName{Name: apicast.DeploymentName, Namespace: apicast.Namespace}, apicastDeployment)
```
miguelsorianod: I understand the way to update status fields is by going to look at the source of truth (in this case the Deployment) instead of focusing on the CR contents.

pkg/controller/apicast/apicast_controller.go
```diff
@@ -189,6 +191,41 @@ func (r *ReconcileAPIcast) Reconcile(request reconcile.Request) (reconcile.Resul
...
+	if err != nil && !errors.IsNotFound(err) {
+		return reconcile.Result{}, err
+	}
+
+	if err != nil && errors.IsNotFound(err) {
+		// TODO should we return an error to requeue or should we stop update
+		// execution at this point and end normally?
+		return reconcile.Result{}, err
```
miguelsorianod: What should we do if for some reason the API returns that the Deployment does not exist yet?

pkg/controller/apicast/apicast_logic_reconciler.go
```diff
@@ -286,6 +286,38 @@ func (r *APIcastLogicReconciler) UserProvidedSecretResourceVersionAnnotations(us
 	return annotations
 }
 
+func (r *APIcastLogicReconciler) APIcastFromCRContents() (*apicast.APIcast, error) {
```
miguelsorianod: This method has been implemented in order to be able to obtain an APIcast instance from outside the reconciler code.
Notice how we get the user provided secrets (and just get, not reconcile them like we do in the Reconcile method)

pkg/controller/apicast/apicast_controller.go
```diff
@@ -189,6 +191,41 @@ func (r *ReconcileAPIcast) Reconcile(request reconcile.Request) (reconcile.Resul
 	return reconcile.Result{}, nil
 }
 
+func (r *ReconcileAPIcast) updateStatus(instance *appsv1alpha1.APIcast, reconciler *APIcastLogicReconciler) (reconcile.Result, error) {
```
miguelsorianod: We've implemented the updateStatus method here in the apicast_controller code.
We could make a public method ReconcileStatus or UpdateStatus inside the APIcastLogicReconciler or other alternative maybe. Maybe it's good enough here for now.

miguelsorianod:
Ready for review @eguzki

pkg/apis/apps/v1alpha1/apicast_types.go

pkg/controller/apicast/apicast_controller.go
eguzki: returning the error to re-queue is good enough

pkg/controller/apicast/apicast_logic_reconciler.go
eguzki: this sounds like a function comment, not a PR comment

miguelsorianod:
> this sounds like a function comment, not a PR comment

I'll include it as a function comment :+1: 

> Since you need to create an instance of apicast.Apicast to get DeploymentName, it would be much easier and less code to implement status reconciliation at the end of apicast.APIcastLogicReconciler.Reconcile() function

About this, I thought that same approach at the beginning but conceptuall at this point it seems separate logic from from the Reconcile function that reconciles the logic of APIcast. In some parts we might want to just update the status without reconciling APIcast (or viceversa)

miguelsorianod:
PR updated to include the clarifying comment on the APIcastfromCR method and removed clarified TODO comments

pkg/controller/apicast/apicast_controller.go
```diff
@@ -189,6 +188,34 @@ func (r *ReconcileAPIcast) Reconcile(request reconcile.Request) (reconcile.Resul
...
+	if err != nil && !errors.IsNotFound(err) {
+		return reconcile.Result{}, err
+	}
+
+	if err != nil && errors.IsNotFound(err) {
+		return reconcile.Result{}, err
+	}
+
+	deployedImage := apicastDeployment.Spec.Template.Spec.Containers[0].Image
+	if !reflect.DeepEqual(instance.Status.Image, deployedImage) {
```
eguzki: Compare two strings with `reflect.DeepEqual` is like "kiling flies with cannons"?

pkg/controller/apicast/apicast_controller.go
```diff
@@ -189,6 +188,34 @@ func (r *ReconcileAPIcast) Reconcile(request reconcile.Request) (reconcile.Resul
 	return reconcile.Result{}, nil
 }
 
+func (r *ReconcileAPIcast) updateStatus(instance *appsv1alpha1.APIcast, reconciler *APIcastLogicReconciler) (reconcile.Result, error) {
```
eguzki: Where is this function called?

pkg/controller/apicast/apicast_controller.go
miguelsorianod: It can be changed to simple comparison yes.

pkg/controller/apicast/apicast_controller.go
miguelsorianod: Fixed.

Output when deploying APIcast with the operator:

```
msoriano@localhost:~/go/src/github.com/3scale/apicast-operator/example-manifests (add-image-to-status)$ oc get apicast example-apicast -o yaml
kind: APIcast
metadata:
  annotations:
    apicast.apps.3scale.net/operator-version: 0.1.0
  creationTimestamp: 2020-01-22T15:39:38Z
  generation: 2
  name: example-apicast
  namespace: msoriano-test
  resourceVersion: "178001"
  selfLink: /apis/apps.3scale.net/v1alpha1/namespaces/msoriano-test/apicasts/example-apicast
  uid: 663f96ed-3d2d-11ea-b510-0800271e62c0
spec:
    name: apicast-gateway-configuration
  replicas: 1
status:
  image: quay.io/3scale/apicast:nightly
```

If I overwrite the image field in spec the image field in status gets updated accordingly:

```
msoriano@localhost:~/go/src/github.com/3scale/apicast-operator/example-manifests (add-image-to-status)$ oc get apicast example-apicast -o yaml
kind: APIcast
metadata:
  annotations:
    apicast.apps.3scale.net/operator-version: 0.1.0
  creationTimestamp: 2020-01-22T15:39:38Z
  generation: 3
  name: example-apicast
  namespace: msoriano-test
  resourceVersion: "178922"
  selfLink: /apis/apps.3scale.net/v1alpha1/namespaces/msoriano-test/apicasts/example-apicast
  uid: 663f96ed-3d2d-11ea-b510-0800271e62c0
spec:
    name: apicast-gateway-configuration
  image: quay.io/3scale/apicast:v3.7.0
  replicas: 1
status:
  image: quay.io/3scale/apicast:v3.7.0
```

miguelsorianod:
Ready for review again @eguzki

pkg/controller/apicast/apicast_controller.go
eguzki: Makes sense that status reconciliation is implemented outside `APIcastLogicReconciler`. So maybe, best to go is making `reconciler.APIcastFromCRContents()` helper method to be shared. Can be future improvement

PR #57, opened 2020-03-25
e2e test with apicast deployed exposing an ingress

eguzki:


eguzki:
It is expected to fail when running on OCP. That is this test is about. 

When rebased on top of the fix https://github.com/3scale/apicast-operator/pull/56 all tests should pass

thomasmaas:
Just curious, could this potentially be a very slow test?

miguelsorianod:
> Just curious, could this potentially be a very slow test?

What this test is specifically checking is not very slow. What is slow is what all e2e tests share, which is the environment set up time: This is, deploying the K8s/OpenShift cluster, deploying the operator, ...

eguzki:
Now, all the tests are expected to pass

PR #69, opened 2020-04-01
Reconciliation refactor: options provider and generic reconciler

eguzki:
* Options provider with validations
* Delegate validation logic to https://godoc.org/github.com/go-playground/validator
* Generic reconciler for all k8s resources. Includes Delete reconciliation logic. 
  * `reconcilers` package with all base reconcilers
* Fix Ingress deletion when `exposedHost` is deleted from CR
* Several other small fixes:
  * Makefile `download` task to populate dependencies
  * User secret management refactor. Secrets are now part of apicast options.
  * apicast_controller: do not show stacktrace when requeued.


eguzki:

pkg/apicast/apicast.go
```diff
@@ -226,25 +217,26 @@ func (a *APIcast) Deployment() *appsv1.Deployment {
...
 				},
 			},
-			Replicas: &a.Replicas, // TODO set to nil?
+			Replicas: &a.options.Replicas,
 		},
 	}
-	if a.OwnerReference != nil {
-		addOwnerRefToObject(deployment, *a.OwnerReference)
+
+	if err := controllerutil.SetControllerReference(a.options.Owner, deployment, a.scheme); err != nil {
```
miguelsorianod, line 224: On the original design we checked whether ownerReference was nil or not (and allowed it to be) because in the design of the APIcast class we wanted to allow defining an APIcast without an owner.
Do you think it's not worth it?

pkg/apicast/apicast.go
```diff
@@ -169,16 +161,15 @@ func (a *APIcast) deploymentEnv() []v1.EnvVar {
 	return env
 }
 
-func (a *APIcast) Deployment() *appsv1.Deployment {
-
+func (a *APIcast) Deployment() (*appsv1.Deployment, error) {
```
miguelsorianod, line 164: We allow returning errors in the returning of objects model?

This seems a conceptual change on how we handle the data model objects.
Are we going to allow errors on it from now on? If so, this change applies to 3scale-operator too and its models?

pkg/k8sutils/object.go
```diff
@@ -7,11 +7,45 @@ import (
...
+	// Add custom annotation
+	annotations := obj.GetAnnotations()
+	if annotations == nil {
+		annotations = make(map[string]string)
+		obj.SetAnnotations(annotations)
+	}
+	annotations[DeleteTagAnnotation] = "true"
+}
+
+func IsObjectTaggedTorDelete(obj KubernetesObject) bool {
```
miguelsorianod, line 43: typo?

pkg/reconcilers/base_reconciler.go
```diff
@@ -0,0 +1,113 @@
...
+	if err != nil {
+		return err
+	}
+
+	key, err := client.ObjectKeyFromObject(existing)
+	if err != nil {
+		return err
+	}
+
+	if err = b.Client().Get(context.TODO(), key, existing); err != nil {
```
miguelsorianod, line 69: I'm trying to understand the logic of this method.

Why do we copy the `desired` object to the `existing` variable and then we perform a Get on the `existing` variable instead of initializing the `existing` object to an empty object?

Apart from that, in case the object exists  when performing the Get, are we sure no contents are "merged"?

pkg/reconcilers/base_reconciler.go
```diff
@@ -0,0 +1,113 @@
...
+
+func (b *BaseReconciler) Scheme() *runtime.Scheme {
+	return b.scheme
+}
+
+func (b *BaseReconciler) Logger() logr.Logger {
+	return b.logger
+}
+
+func (b *BaseReconciler) ReconcileResource(desired k8sutils.KubernetesObject, mutateFn MutateFn) error {
```
miguelsorianod, line 58: I can't find a reference to this ReconcileResource method. Shouldn't be used somewhere?

pkg/apicast/apicast.go
eguzki: I did not see any value making it optional. No event or condition leads to ownerless objects. If you think it is worth for any reason, let me know and I can make it optional.

pkg/apicast/apicast.go
eguzki: The only source of error is the method `controllerutil.SetControllerReference` because it checks if the objects already has a owner. If it already has and it is different to the one trying to be set, error is raised. 

Assuming the objects do not have owners, because they are being created, it is safe to assign owner the way it was before:
```
o.SetOwnerReferences(append(o.GetOwnerReferences(), owner))
```
I will change it

pkg/k8sutils/object.go
eguzki: :+1:

pkg/apicast/apicast.go
eguzki: I remember the motivation behind doing this way: I did not want to implement this function:
```
// asOwner returns an owner reference set as the tenant CR         
func asOwner(a *appsv1alpha1.APIcast) metav1.OwnerReference {      
    trueVar := true                                                
    return metav1.OwnerReference{                                  
        APIVersion: appsv1alpha1.SchemeGroupVersion.String(),      
        Kind:       appscommon.APIcastKind,                        
        Name:       a.Name,                                        
        UID:        a.UID,                                         
        Controller: &trueVar,                                      
    }                                                              
}                                                                  
```
This OwnerReference object building is done by `controllerutil.SetControllerReference` method and I wanted to use it instead of implementing it. The toll was having errors returned. If you prefer implementing it, no problem.

pkg/reconcilers/base_reconciler.go
eguzki: `existing` object cannot be initialized to an empty object because you need an object of some specific `Kind`. As I understand, k8s client `Get` method accepts `runtime.Object`, but needs `schema.ObjectKind` set correctly in order to de-serialize properly. Thus, `existing` is created by callling the desired object's `runtime.Object.DeepCopyObject()` that preserves the type of object. 


Regarding your second question. I am assuming that the `existing` object will represent accurately the k8s resource existing in the cluster. Thus, properties will be initialized and  values set from the k8s resource serialization in the response of the request. wdyt?

pkg/apicast/apicast.go
eguzki: changed

pkg/apicast/apicast.go
miguelsorianod: Ok. Lets leave it this way. We can always change it in the future again if it was needed, which does not seem so for the moment

pkg/apicast/apicast.go
miguelsorianod: Just to clarify, I was not against changing it if you prefer the new way you implemented it.
I was just asking whether we changed the way of dealing with it (which might be the appropriate one).
I see you've changed it already but if you prefer the previous way I'm ok with it too if it means we in general would be using the new approach

pkg/reconcilers/base_reconciler.go
miguelsorianod: I get the answer to your first question. It makes sense :+1:

miguelsorianod: > Regarding your second question. I am assuming that the existing object will represent accurately the k8s resource existing in the cluster. Thus, properties will be initialized and values set from the k8s resource serialization in the response of the request. wdyt?

I think I understand what you say. You are telling me that the values of existing would be overwritten on top of the  ones that have been defined in the desired. My concern was about the ones that might not be defined too.

To try to give you an example, the situation would be:
* There's a field defined in the desired object but the real existing object does not define it
What would it happen with the contents of that field? would the variable would end up containing the field?

pkg/reconcilers/base_reconciler.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/go-logr/logr"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/3scale/apicast-operator/pkg/k8sutils"
+)
+
+// MutateFn is a function which mutates the existing object into it's desired state.
+type MutateFn func(existing, desired k8sutils.KubernetesObject) (bool, error)
```
miguelsorianod, line 16: what MutateFn takes is a k8sutils.KubernetesObject.
It usually happens that you are interested on specific fields of a specific Kind.

I understand then that the idea (as I've seen you've done in some of them) is that the MutateFn methods that we implement should contain a type assertion to check and convert to the desired type right?

pkg/controller/apicast/apicast_logic_reconciler.go
```diff
@@ -403,144 +128,110 @@ func (r *APIcastLogicReconciler) applyInitialization() bool {
...
-	existingService := v1.Service{}
-	err := r.Client().Get(context.TODO(), r.namespacedName(&desiredService), &existingService)
-	if err != nil {
-		if errors.IsNotFound(err) {
-			r.Logger().Info(fmt.Sprintf("Creating %s", k8sutils.ObjectInfo(&desiredService)))
-			err = r.Client().Create(context.TODO(), &desiredService)
-		}
-		return err
+func IngressMutator(existingObj, desiredObj k8sutils.KubernetesObject) (bool, error) {
+	existing, ok := existingObj.(*extensions.Ingress)
```
miguelsorianod, line 207: I'm curious about the effects of the type assertion/conversion

I thought type convertion/assertion performs a copy of the object that it tries to convert.

If you make changes, wouldn't they be made to the copy (`existing`) and thus not be effective in `existingObj` when going out of the method?

miguelsorianod:
Resolved some comments and left some more feedback

pkg/reconcilers/base_reconciler.go
eguzki: I have checked it with the following test:

Create apimanager object:

```
$ k get apimanager apimanager1 -o yaml
kind: APIManager
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"apps.3scale.net/v1alpha1","kind":"APIManager","metadata":{"annotations":{},"name":"apimanager1","namespace":"eguzki-metrics"},"spec":{"resourceRequirementsEnabled":false,"wildcardDomain":"eguzki-test.apps.dev-eng-operator-ocp4-2.dev.3sca.net","zync":{"image":"existingImage"}}}
  creationTimestamp: "2020-04-09T10:26:53Z"
  generation: 1
  name: apimanager1
  namespace: eguzki-metrics
  resourceVersion: "19104905"
  selfLink: /apis/apps.3scale.net/v1alpha1/namespaces/eguzki-metrics/apimanagers/apimanager1
  uid: a15c8bc9-7a4c-11ea-b0df-12d443f7f383
spec:
  resourceRequirementsEnabled: false
  wildcardDomain: eguzki-test.apps.dev-eng-operator-ocp4-2.dev.3sca.net
  zync:
    image: existingImage
```

Note the existing fields, `wildcardDomain`, `zync.image`.

Then, I will create a APIManager object:

```
tmpTenant := "mytenantname"                                       
tmpzyncImage := "desiredZyncImage"                                
tmpSystemImage := "desiredSystemImage"                            
desiredAPIManager := &v1alpha1.APIManager{                        
    Spec: v1alpha1.APIManagerSpec{                                
        APIManagerCommonSpec: v1alpha1.APIManagerCommonSpec{      
            TenantName: &tmpTenant,                               
        },                                                        
        Zync: &v1alpha1.ZyncSpec{                                 
            Image: &tmpzyncImage,                                 
        },                                                        
        System: &v1alpha1.SystemSpec{                             
            Image: &tmpSystemImage,                               
        },                                                        
    },                                                            
}                                                                                                                                  
```
Note:
* TenantName new field
* Wildcard property not defined
* Zync.Image property defined and different from existing one
* System.image propoerty new field.

So we have:
* In the existing object:
  * properties not in the desired object: wildcardDomain
  * properties in the desired object: zync.image
* In the desiredObject:
   * properties not in the existing object: tenantName, system.image
   * properties in the existing object: zync.image

The I will run this simple code:

```
fmt.Println("BEFORE")                                                   
s, _ := json.MarshalIndent(desiredAPIManager, "", "\t")                 
fmt.Println(string(s))                                                  
if err := cl.Get(context.TODO(), key, desiredAPIManager); err != nil {  
    panic(err)                                                          
}                                                                       
fmt.Println("AFTER")                                                    
s, _ = json.MarshalIndent(desiredAPIManager, "", "\t")                  
fmt.Println(string(s))                                                  
              
```

And the output is this:

```
BEFORE
{
	"metadata": {
		"creationTimestamp": null
	},
	"spec": {
		"wildcardDomain": "",
		"tenantName": "mytenantname",
		"system": {
			"image": "desiredSystemImage"
		},
		"zync": {
			"image": "desiredZyncImage"
		}
	},
	"status": {
		"deployments": {}
	}
}

AFTER

{
	"metadata": {
		"name": "apimanager1",
		"namespace": "eguzki-metrics",
		"selfLink": "/apis/apps.3scale.net/v1alpha1/namespaces/eguzki-metrics/apimanagers/apimanager1",
		"uid": "a15c8bc9-7a4c-11ea-b0df-12d443f7f383",
		"resourceVersion": "19104905",
		"generation": 1,
		"creationTimestamp": "2020-04-09T10:26:53Z",
		"annotations": {
			"kubectl.kubernetes.io/last-applied-configuration": "{\"apiVersion\":\"apps.3scale.net/v1alpha1\",\"kind\":\"APIManager\",\"metadata\":{\"annotations\":{},\"name\":\"apimanager1\",\"namespace\":\"eguzki-metrics\"},\"spec\":{\"resourceRequirementsEnabled\":false,\"wildcardDomain\":\"eguzki-test.apps.dev-eng-operator-ocp4-2.dev.3sca.net\",\"zync\":{\"image\":\"existingImage\"}}}\n"
		}
	},
	"spec": {
		"wildcardDomain": "eguzki-test.apps.dev-eng-operator-ocp4-2.dev.3sca.net",
		"tenantName": "mytenantname",
		"resourceRequirementsEnabled": false,
		"system": {
			"image": "desiredSystemImage"
		},
		"zync": {
			"image": "existingImage"
		}
	},
	"status": {
		"deployments": {}
	}
}
```

Conclusion:
* Properties in the existing object and not in the desired object, wilcarddomain: updated correctly in desired object [OK]
* Properties not in the existing object and in the desired object, tenant name and system.image: Exist in the desired object after `Get` call. [NOT OK]
* Properties in the existing object and in the desired object: zync.image: updated correctly in desired object [OK]

So, I guess this is a problem because attributes in the desired object not existing in the cluster will still be there after reading from the cluster. I will change the signature of the function from
```
func (b *BaseReconciler) ReconcileResource(desired k8sutils.KubernetesObject, mutateFn MutateFn) error
```

to 

```
func (b *BaseReconciler) ReconcileResource(obj, desired k8sutils.KubernetesObject, mutateFn MutateFn) error
```
And it is responsability of the callee to pass "empty" object with a correct Kind.

Something like this:

```
obj := &appsv1.Deployment{}
err = r.ReconcileResource(obj, desiredDeployment, DeploymentMutator)
```

pkg/reconcilers/base_reconciler.go
eguzki: Yes. It is a toll to be paid. There is no other way, AFAIK. Golang requires it.

miguelsorianod: Yes that's what I thought. Just wanted to make sure I understood it for future implementations of those methods :+1:

pkg/reconcilers/base_reconciler.go
miguelsorianod: :+1: 

Thank you for looking into this and trying it.

That seems to be the solution yes. It's not as clean but it's how k8s implements deserialization in the receiving parameter in the Get so not much we can do about that I think unfortunately.

pkg/controller/apicast/apicast_logic_reconciler.go
eguzki: If you run this snippet:

```
import (                                                    
    "encoding/json"                                         
    "fmt"                                                   
)                                                           
                                                            
type I interface {                                          
    Walk()                                                  
}                                                           
                                                            
type S struct {                                             
    B int                                                   
}                                                           
                                                            
func (s *S) Walk() {                                        
}                                                           
                                                            
func updateI(i I) {                                         
    newS, ok := i.(*S)                                      
    if !ok {                                                
        panic("not ok")                                     
    }                                                       
    newS.B = 6                                              
}                                                           
                                                            
func print(s *S) {                                          
    se, _ := json.MarshalIndent(s, "", "\t")                
    fmt.Println(string(se))                                 
}                                                           
                                                            
func main() {                                               
    s := &S{B: 5}                                           
                                                            
    fmt.Println("before")                                   
    print(s)                                                
                                                            
    updateI(s)                                              
                                                            
    fmt.Println("after")                                    
    print(s)                                                
}                                                           

```

The output is:

```
before
{
	"B": 5
}
after
{
	"B": 6
}
```

This is easy to explain. Asserted type is pointer type.

pkg/reconcilers/base_reconciler.go
eguzki: implemented that way

pkg/reconcilers/base_reconciler.go
```diff
@@ -0,0 +1,108 @@
...
+
+func (b *BaseReconciler) Scheme() *runtime.Scheme {
+	return b.scheme
+}
+
+func (b *BaseReconciler) Logger() logr.Logger {
+	return b.logger
+}
+
+func (b *BaseReconciler) ReconcileResource(existing, desired k8sutils.KubernetesObject, mutateFn MutateFn) error {
```
miguelsorianod, line 58: Can we add some comment to explain what is the expected value of `existing`  and what's the behaviour in relation to it?

It would help clarify the behaviour of the method for users of it

pkg/reconcilers/base_reconciler.go
miguelsorianod: :+1:

pkg/controller/apicast/apicast_logic_reconciler.go
miguelsorianod: Ok, so I understand then that the method expects an interface, and the interface contains a value that's a pointer to an object.

This means that what is copied when doing the assertion is the pointer value (and not what the pointer points to).

This implies that for example you could not replace the entire object itself and get the new object outside of the method because the pointer is copied but you can change the fields of the object :+1:

pkg/reconcilers/base_reconciler.go
eguzki: :+1:

eguzki: Some rewording done. Review required @miguelsorianod

miguelsorianod: :+1:

PR #79, opened 2020-09-17
Resource requirements

eguzki:
This PR adds configurable resource requirements


pkg/apicast/apicast_options.go
```diff
@@ -46,3 +48,16 @@ func (a *APIcastOptions) Validate() error {
 	validate := validator.New()
 	return validate.Struct(a)
 }
+
+func DefaultResourceRequirements() v1.ResourceRequirements {
```
miguelsorianod, line 52: This seems it is not being use anywhere. I understand we should use this and set it as the apicast's resourcerequirements when the user has not specified resources in the CR

miguelsorianod (changes requested):

miguelsorianod (approved):
:+1:

PR #80, opened 2020-09-18
handle update conflict error

eguzki:


thomasmaas:
Any related JIRA for this @eguzki ?

miguelsorianod (approved):

eguzki:
@thomasmaas small enhancement to handle conlict updates (when the operator tries to update a resource from an outdated representation) instead to treat them as cluster errors. In both types or error, operator is retried, so not other effect than better logging, avoiding ugly errors that are not important errors.

PR #81, opened 2020-09-21
contextual logging

eguzki:
Apicast operator supports multiple apicast instances. This PR adds  context about the apicast instance the log related to (CR name and namespace)

Note that the baseController's logger instance will be contextualized to the apicast instance.


eguzki:
ping @miguelsorianod

miguelsorianod (approved):

PR #86, opened 2020-10-13
Add disconnected install OpenShift annotation

miguelsorianod:


miguelsorianod:
Yes, I think so. It applies to all releases from now on.

PR #96, opened 2020-11-06
Enable TLS at pod level

eguzki:


apis/apps/v1alpha1/apicast_types.go
```diff
@@ -181,6 +192,26 @@ func (a *APIcast) GetOwnerRefence() *metav1.OwnerReference {
 
 func (a *APIcast) Reset() { *a = APIcast{} }
 
+func (a *APIcast) Validate() field.ErrorList {
+	errors := field.ErrorList{}
+
+	// check HTTPSPort is required for httpsCertificateSecretRef
```
miguelsorianod, line 198: Should using a default port be a good idea?

miguelsorianod: What does trigger enablement of HTTPs in APIcast? specifying the port? or the certifcate?

apis/apps/v1alpha1/apicast_types.go
```diff
@@ -92,6 +93,16 @@ type APIcastSpec struct {
 	// ServiceConfigurationVersionOverride contains service configuration version map to prevent it from auto-updating.
 	// +optional
 	ServiceConfigurationVersionOverride map[string]string `json:"serviceConfigurationVersionOverride,omitempty"` // APICAST_SERVICE_${ID}_CONFIGURATION_VERSION
+	// HttpsPort controls on which port APIcast should start listening for HTTPS connections. If this clashes with HTTP port it will be used only for HTTPS.
+	// +optional
+	HTTPSPort *int32 `json:"httpsPort,omitempty"` // APICAST_HTTPS_PORT
+	// HTTPSVerifyDepth defines the maximum length of the client certificate chain.
+	// +kubebuilder:validation:Minimum=0
+	// +optional
+	HTTPSVerifyDepth *int `json:"httpsVerifyDepth,omitempty"` // APICAST_HTTPS_VERIFY_DEPTH
```
miguelsorianod, line 102: This should be `*int32` (or *int64) because int is ambiguos depending on the platform, as per API Conventions (see https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md, "primitive types" subsection)
I'd go for *int64, and also for HTTPSPort, to be consistent with the other defined fields in the CR to not have a mix of sizes (even if not that much length is required).

eguzki: Ok, changing to `*int64`. 

HTTPSPort is better to be `*int32` as it is the type required by `corev1.ContainerPort` type's `ContainerPort` field

apis/apps/v1alpha1/apicast_types.go
eguzki: I guess this is a decision up to us. The APIcast will not activate https unless HTTPSPort is provided in env var.

The current implementation ensures both (certificate and https port) are provided or none. Another approach is that if the customer provides the certificate and does not provide the port, the operator takes a default one on behalf of the customer (for instance 443 or 8081). I think this approach requires less config from the customer. I will implement.

eguzki:
There is an issue when the HTTPSPort conflicts with the HTTP port and the operator deploys an ingress. If the CR does not provide certificate secret, the ingress will be HTTP only and if the CR provides certficate secret, the ingress will be HTTPS edge. In any case, the traffic to the apicast service will be HTTP  based and APIcast will only serve HTTPS traffic, therefore deployment issue as traffic will not be served (If I am not wrong).

Another option is to deploy ingress with `SSL passthrough`, but this is done via annotation and it is ingress controller specific. Openshift deplos, by default HAproxy controller, but the cluster can have other controllers, like nginx based ingress controlled. Does not seem to be a solution to cover all cases. 

What should the operator do when the ports conflict? One option is declare the CR as invalid and request customer intervention. 

wdyt @miguelsorianod ?

eguzki:
@eloycoto you may want to update this doc https://github.com/3scale/APIcast/blob/master/doc/parameters.md#apicast_https_port about `If this clashes with HTTP port it will be used only for HTTPS.`. After trying it, if there is a clash, the apicast pod does not start with the error:
```
nginx: [emerg] a duplicate listen 0.0.0.0:8080 in /tmp/lua_dAb3TD:349
```

apis/apps/v1alpha1/apicast_types.go
eguzki: implemented default port approach. Ready for review.

doc/apicast-crd-reference.md
```diff
@@ -33,6 +33,7 @@
 | `loadServicesWhenNeeded` | bool | No | false | The configurations are loaded lazily (see [docs](https://github.com/3scale/APIcast/blob/master/doc/parameters.md#apicast_load_services_when_needed)) |
 | `servicesFilterByURL` | string | No | N/A |  Used to filter the service configured in the 3scale API Manager, the filter matches with the public base URL (Staging or production) (see [docs](https://github.com/3scale/APIcast/blob/master/doc/parameters.md#apicast_services_filter_by_url)) |
 | `serviceConfigurationVersionOverride` | [Service Configuration Version Override object](#service-configuration-version-override-map) | No | N/A | Service configuration version map to prevent it from auto-updating (see [docs](https://github.com/3scale/APIcast/blob/master/doc/parameters.md#apicast_service_id_configuration_version)) |
+| `httpsPort` | int | No | N/A | Controls on which port APIcast should start listening for HTTPS connections. If this clashes with HTTP port it will be used only for HTTPS. (see [docs](https://github.com/3scale/APIcast/blob/master/doc/parameters.md#apicast_https_port)) |
```
miguelsorianod, line 36: Based on what I read I think this is outdated right? If it clashes with http port an error will be reported at CR level

apis/apps/v1alpha1/apicast_types.go
miguelsorianod: :+1:

apis/apps/v1alpha1/apicast_types.go
miguelsorianod: :+1:

miguelsorianod (approved):

PR #126, opened 2021-06-15
Custom policies added to the CRD

eguzki:
https://issues.redhat.com/browse/THREESCALE-7031

Custom policies exposed in the CRD:

```yaml
kind: APIcast                              
metadata:                                  
  name: apicast1                           
spec:                                      
  customPolicies:                          
    - secretRef:
        name: cp-1                     
      name: Example                        
      version: "0.1"                       
    - secretRef:
        name: cp-1    
      name: Example                        
      version: "0.2"            
   - secretRef:
        name: cp-3                 
     name: OtherExample                        
     version: "0.1"                     
```

Each policy is defined by:
* Name
* Version
* Secret containing at least: "init.lua" and "apicast-policy.json" keys. Other lua files can also be included.
```
apiVersion: v1
data:
  apicast-policy.json:   CONTENT OF THE POLICY METADATA (required)
  init.lua: CODE OF THE POLICY (required)
  example.lua: CODE OF THE POLICY (optional)
  other.lua: CODE OF THE POLICY (optional)
kind: Secret
type: Opaque
```

All fields are mandatory.

Related liniks:
* https://github.com/3scale/APIcast/blob/master/doc/policies.md#write-your-own-policy


apis/apps/v1alpha1/apicast_types.go
```diff
@@ -28,6 +30,21 @@ import (
...
 
+// CustomPolicySpec contains or has reference to an APIcast custom policy
+type CustomPolicySpec struct {
+	// Name specifies the name of the custom policy
+	Name string `json:"name"`
+	// Version specifies the name of the custom policy
+	Version string `json:"version"`
+
+	// SecretName specifies the secret name holding the custom policy metadata and lua code
+	SecretName string `json:"secretName"`
```
miguelsorianod, line 41: If we are going to reference a secret and always in the same namespace, can we use v1.LocalObjectReference as with other fields like HTTPSCertificateSecretRef and AdminPortalCredentialsRef?

apis/apps/v1alpha1/apicast_types.go
```diff
@@ -28,6 +30,21 @@ import (
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
+// CustomPolicySpec contains or has reference to an APIcast custom policy
+type CustomPolicySpec struct {
```
miguelsorianod, line 34: If in the future we think we might implement other ways to provide the policy maybe it would make sense to have a section for it like:

customPolicySpec:
  name:
  version:
  policySource:
    secret:
      secretRef:
        ...
    inline:
      relevantfieldToInlinePolicy1
      relevantfieldToInlinePolicy2
      ...
    ...
```

What do you think?

File: pkg/apicast/apicast.go [hunk 3]
```diff
@@ -67,9 +71,21 @@ func (a *APIcast) deploymentVolumeMounts() []v1.VolumeMount {
...
+			MountPath: path.Join(CustomPoliciesMountBasePath, customPolicySpec.Name, customPolicySpec.Version),
+			ReadOnly:  true,
+		})
+	}
+
 	return volumeMounts
 }
 
+func policyVolumeName(spec appsv1alpha1.CustomPolicySpec) string {
+	return fmt.Sprintf("%s-%s", "policy", helper.DNS1123Name(spec.VersionName()))
```
miguelsorianod, line 86: I am curious about this. What's the limitation here? do you have some link to the doc for the limitation?

pkg/apicast/apicast.go
```diff
@@ -67,9 +71,21 @@ func (a *APIcast) deploymentVolumeMounts() []v1.VolumeMount {
 		})
 	}
 
+	for _, customPolicySpec := range a.options.CustomPolicies {
+		volumeMounts = append(volumeMounts, v1.VolumeMount{
```
miguelsorianod, line 75: I'm curious about this behavior.
What would happen if the user updates the secret content? Is that immediately visible in the container's filesystem?
I'm guessing that independently on that apicast might or might not be able to react to file changes once it's running. Do you know about that too?

pkg/apicast/apicast.go
```diff
@@ -67,9 +71,21 @@ func (a *APIcast) deploymentVolumeMounts() []v1.VolumeMount {
...
+			Name:      policyVolumeName(customPolicySpec),
+			MountPath: path.Join(CustomPoliciesMountBasePath, customPolicySpec.Name, customPolicySpec.Version),
```
miguelsorianod, line 77: Are the directories automatically created if they not exist?

apis/apps/v1alpha1/apicast_types.go
eguzki: `Name` and `Version` are mandatory in any case. 

I agree, The secret (v1.LocalObjectReference type) can be optional (but make it mandatory in this implementation) and let in the future add some `inline` object.

pkg/apicast/apicast.go
eguzki: The volume name does not accept dots, `.`. Only DNS1123 compliant names. When not sanitized, I had an error.

```
// Volume represents a named volume in a pod that may be accessed by any container in the pod.   
type Volume struct {                                                                             
    // Volume's name.                                                                            
    // Must be a DNS_LABEL and unique within the pod.                                            
    // More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names 
    Name string `json:"name" protobuf:"bytes,1,opt,name=name"`                                   
```

Adding `policy` prefix to know more context about the volume when you list volumes of the Deployment resource.

pkg/apicast/apicast_options.go
```diff
@@ -50,6 +52,7 @@ type APIcastOptions struct {
 	HTTPSCertificateSecret              *v1.Secret
 	Workers                             *int32
 	Timezone                            *string
+	CustomPolicies                      []appsv1alpha1.CustomPolicySpec
```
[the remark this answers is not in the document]
eguzki: Ok, but then I will have to duplicate the type. That would be ok for you?

pkg/apicast/apicast.go
eguzki: I will do some testing about this. My guess, apicast will not notice the change even if the new content is available in the volume.

If updating the content of the secret does not automatically make it available to the container (which I do not think so):
* Watch for secret updates and redeploy
* Document that secret updates are not visible to the apicast and you need to redeploy after secret has been updated. Or create a new secret and change reference in the CR. 

For the first iteration, I would go for the doc solution. wdyt?

pkg/apicast/apicast.go
eguzki: yup

pkg/apicast/apicast_options.go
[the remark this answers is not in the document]
miguelsorianod: Yes. That would be ok :+1:

pkg/apicast/apicast.go
miguelsorianod: seems reasonable to me :+1:

pkg/apicast/apicast.go
miguelsorianod: :+1: thanks for the explanation

eguzki:
When the mounted secret is updated externally, the container can automatically see the changes. However, apicast has the policy already loaded and does not change behavior. When redeployed, updating the `replica` field, the policy update is applied.

apis/apps/v1alpha1/apicast_types.go
eguzki: updated

apis/apps/v1alpha1/apicast_types.go
eguzki: updated

pkg/apicast/apicast_options.go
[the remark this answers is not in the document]
eguzki: updated

pkg/apicast/apicast.go
miguelsorianod: :+1:

pkg/apicast/apicast_option_provider.go
```diff
@@ -122,6 +122,15 @@ func (a *APIcastOptionsProvider) GetApicastOptions() (*APIcastOptions, error) {
 	a.APIcastOptions.Workers = a.APIcastCR.Spec.Workers
 	a.APIcastOptions.Timezone = a.APIcastCR.Spec.Timezone
 
+	for _, customPolicySpec := range a.APIcastCR.Spec.CustomPolicies {
```
miguelsorianod, line 125: I'm wondering whether we should validate each secret first, when processing the options?
In the case of HTTPsCertificate we check that it exists beforehand(see `getHTTPSCertificateSecret` method). Maybe check that the specific required fields in the secret are there too?

What do you think?

apis/apps/v1alpha1/apicast_types.go
miguelsorianod: :+1: .

Do you think it might end up being confusing if we end up having multiple common attributes together with the different policySource types (a union) at the same level? imagine there could be other unions at the same level too.

The current implementation would end up looking as:
```
customPolicySpec:
  name:
  version:
  commonAttr1:
  commonAttr2:
  secretRef:
  inline:
    ...
  anotherPolicySourceType:
    ...
  aNewUnionTypeType1:
    ...
  aNewUnionTypeType2:
    ...
  ...
```

vs what it was proposed in the initial comment:

```
customPolicySpec:
  name:
  version:
  commonAttr1:
  commonAttr2:
  policySource:
    secret:
      secretRef:
      ...
    inline:
      relevantfieldToInlinePolicy1:
      relevantfieldToInlinePolicy2:
      ...
    ...
  aNewUnionType:
    aNewUnionTypeType1:
      ...
    anewUnionTypeType2:
```

The intention of the proposal in my previous comment was to try to avoid this kind of cluttering at the same level of unrelated terms and excluding values, being contained in their own section.

apis/apps/v1alpha1/apicast_types.go
miguelsorianod: :+1:

pkg/apicast/apicast_options.go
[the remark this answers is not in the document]
miguelsorianod: :+1:

pkg/apicast/apicast_option_provider.go
eguzki: yes! additional check would be good

apis/apps/v1alpha1/apicast_types.go
eguzki: I do not think it is necessary or confusing.

There can be other objects at the same level, some of them being policy source types and others common attributes.

miguelsorianod: > There can be other objects at the same level, some of them being policy source types and others common attributes.

I agree. But notice the example I pointed where there would be another equivalent of policy source at the same level.

pkg/apicast/apicast_option_provider.go
eguzki: implemented

apis/apps/v1alpha1/apicast_types.go
eguzki: Sorry, I do not get what would be mutually exclusive. 

aNewUnionTypeType1` is like `commonAttr1` but being an object instead of a simple value. 

```yaml
customPolicySpec:
  - name:
    version:
    commonAttr1:
    commonAttr2:
    secretRef:
    inline:
      ...
    anotherPolicySourceType:
      ...
    aNewUnionTypeType1:
      ...
    aNewUnionTypeType2:
      ...
```

pkg/apicast/apicast.go
eguzki: doc added

apis/apps/v1alpha1/apicast_types.go
miguelsorianod: > Sorry, I do not get what would be mutually exclusive.

Sorry. Let me try to clarify. An example of a "union type", where by definition all of its values are mutually exclusive would be the "policy source". We would have `secret` or `inline`. The user can't / shouldn't not set the two because they are mutually exclusive. They are part of an imaginary "union type". It is also explained a little bit in https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#unions

pkg/apicast/apicast_option_provider.go
```diff
@@ -122,6 +122,26 @@ func (a *APIcastOptionsProvider) GetApicastOptions() (*APIcastOptions, error) {
 	a.APIcastOptions.Workers = a.APIcastCR.Spec.Workers
 	a.APIcastOptions.Timezone = a.APIcastCR.Spec.Timezone
 
+	for idx, customPolicySpec := range a.APIcastCR.Spec.CustomPolicies {
+		namespacedName := types.NamespacedName{
+			Name:      customPolicySpec.SecretRef.Name, // CR Validation ensures not nil
```
miguelsorianod, line 127: Even though CR validation ensures that `customPolicySpec.SecretRef` is not nil, unfortunately it does not ensure that the `Name` is not empty. We should check that name is not empty too. Maybe inside validateCustomPolicySecret method

apis/apps/v1alpha1/apicast_types.go
eguzki: This is a documentation and internal validation issue then. 

When we add the inline policy source, if we ever add it, we need to document that it is mutually exclusive with the secret source and validate in the controller. 

I still think it is ok to go like it is now.

pkg/apicast/apicast_option_provider.go
eguzki: validated

miguelsorianod (approved):
Looks good to me. Great work :+1:

PR #127, opened 2021-06-18
Fix csv alm samples

eguzki:
https://issues.redhat.com/browse/THREESCALE-7226


miguelsorianod (approved):
:+1: . We will need to backport this to 2.11 candidate too

PR #132, opened 2021-06-29
extended metrics

### LINKED ISSUE #131: Add the extended metrics variable to the CRD
at the moment this is only doable with the Template approach for apicast

eguzki:
Fixes #131


apis/apps/v1alpha1/apicast_types.go
```diff
@@ -166,6 +166,11 @@ type APIcastSpec struct {
 	// CustomPolicies specifies an array of defined custome policies to be loaded
 	// +optional
 	CustomPolicies []CustomPolicySpec `json:"customPolicies,omitempty"`
+
+	// ExtendedMetrics enables additional information on Prometheus metrics; some labels will be used with specific information that will provide more in-depth details about APIcast.
```
miguelsorianod, line 170: Does this have an appreciable impact in performance(does it say so in apicast doc?) ? if so maybe it would be a good idea to document it

miguelsorianod (approved):
Looks good to me :+1: I've left a minor comment

apis/apps/v1alpha1/apicast_types.go
eguzki: @eloycoto question for you ^^

PR #141, opened 2021-07-21
Add OpenTracing configurability

miguelsorianod:
This exposes OpenTracing functionality provided by APIcast into the APIcast operator.

### Design

A new `spec.openTracing` section has been added in the `APIcast` CR.

An example:

```
spec:
  openTracing:
    enabled: true
    tracingLibrary: jaeger
    tracingConfigRef:
      name: mysecretname
```

* All attributes in the `openTracing` section are optional
* If the `enabled` flag is not specified, it defaults to `false`
* If the `tracingLibrary` flag is not specified, it defaults to `jaeger`
* If the `tracingConfigRef` flag is not specified, it defaults to a default tracing configuration specific to the set `tracingLirary`. This default tracing configuration is provided in the APIcast image itself

The configuration is mounted into the APIcast pod by looking at a `config` key in the provided secret name and mounted in `/opt/app-root/src/tracing-configs/tracing-config-<tracing_library_name>-<secret_name>`, for example in `opt/app-root/src/tracing-configs/tracing-config-jaeger-mysecretname`

Pending:

- [x] Verification


miguelsorianod:
With the current implementation the OpenTracing configuration is being provided with a `Secret` but in the official APIcast product documentation I see a `ConfigMap` is used. Maybe in this case a `Secret` is not needed and we should use a `ConfigMap` instead?

miguelsorianod:
Ready for a first review @eguzki before I proceed to do some verifications

pkg/apicast/apicast.go
```diff
@@ -151,6 +165,23 @@ func (a *APIcast) deploymentVolumes() []v1.Volume {
 		})
 	}
 
+	if a.options.TracingConfig.TracingConfigSecretName != nil {
```
eguzki, line 168: shouldn't be like that?
```
if a.options.TracingConfig.Enabled && a.options.TracingConfig.TracingConfigSecretName != nil {
```

eguzki (commented):
looks good

pkg/apicast/apicast.go
miguelsorianod: You are right. I will fix it :+1:

miguelsorianod: changed

miguelsorianod:
@eguzki I think this should be ready

pkg/apicast/apicast_option_provider.go
```diff
@@ -338,3 +381,19 @@ func (a *APIcastOptionsProvider) customEnvSecret(nn types.NamespacedName) (*v1.S
...
+	secret := &v1.Secret{}
+	err := a.Client.Get(context.TODO(), nn, secret)
+
+	if err != nil {
+		// NotFoundError is also an error, it is required to exist
+		return err
+	}
+
+	if _, ok := secret.Data[TracingConfigSecretKey]; !ok {
+		return fmt.Errorf("Required secret key, %s not found", "config")
```
eguzki, line 395: minor: use `TracingConfigSecretKey` instead of raw `config`

eguzki (approved):
LGTM

minor comments dropped

apis/apps/v1alpha1/apicast_types.go
```diff
@@ -195,6 +203,22 @@ type APIcastExposedHost struct {
 	TLS []extensions.IngressTLS `json:"tls,omitempty"`
 }
 
+type OpenTracingSpec struct {
+	// Enabled controls whether OpenTracing integration with APIcast is enabled.
+	// By default it is not enabled.
+	// +optional
+	Enabled *bool `json:"enabled,omitempty"`
+	// +optional
```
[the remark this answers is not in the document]
miguelsorianod: Apparently yes because the generated field does not contain required. I will reorder it anyway to be consistent with the other fields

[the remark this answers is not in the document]
miguelsorianod: updated

pkg/apicast/apicast_option_provider.go
miguelsorianod: updated :+1:

apis/apps/v1alpha1/apicast_types.go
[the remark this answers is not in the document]
eguzki: I remember an issue about this, `+optional` is not enough to have it `optional`, the json tag `omitempty` was also required to have the field optional. But maybe it was an old version of operator-sdk or kubebuilder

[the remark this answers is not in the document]
eguzki: in the 3scale operator the mark is also above the comments

PR #142, opened 2021-08-06
fix reconcile service https port

eguzki:
When HTTPS port is removed from the CR, the service's exposed ports were not reconciled. Added reconciliation of the ports.


miguelsorianod:
Looks good to me :+1:

miguelsorianod:
I think we should probably backport this to 2.11-candidate as the original functionality for this was already merged to candidate and this is kind of a relevant change to it

miguelsorianod (approved):

PR #148, opened 2021-09-01
Migrate Ingress to networking v1

miguelsorianod:
Starting from K8s 1.22, the `extensions/v1beta1` and `networking.k8s.io/v1beta1` API versions of Ingress are no longer served (https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-22).

This PR upgrades Ingress to use the `networking.k8s.io/v1` API version to allow apicast-operator to run on K8s >= 1.22

The PR has also required upgrading K8s related libraries to Kubernetes 1.19.x because `networking.k8s.io/v1` was introduced in K8s 1.19.


miguelsorianod:
This PR should be backported to 3scale-2.11-candidate

eguzki (approved):

PR #165, opened 2021-11-22
Delete kube-rbac-proxy container from controller-manager

eguzki:
kube-rbac-proxy is used by default on operator-sdk to protect operator metrics path, in case you don't want anyone in the cluster but only k8s authenticated resources can access to them.

However there are some issues with the kube-rbac-proxy:
* The proxy serves metrics in HTTPS and prometheus needs to be conffigured with TLS skipVerify enabled. The following fix needs to be applied:
```yaml
+      scheme: https
+      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      tlsConfig:
+        insecureSkipVerify: true
```
However **OpenShift User Workload Monitoring** (OpenShift official monitoring stack) does not support ServiceMonitor with bearerTokenFile field [Prometheus ServiceMonitor failing to scrape operator metrics served though kube-proxy HTTPS 8443 port operator-framework/operator-sdk#4764 (comment)](https://github.com/operator-framework/operator-sdk/issues/4764#issuecomment-821264658) (which is needed to scrape metrics behind kube-rbac-proxy) , so it seems there is no way of having operator metrics with auth if using OCP UWM. We are looking at leveraging OCP UWM for a better integration with openshift.

* Prometheus instances require extra permissions to scrape, this is the required cluster scope role to scrape operator metrics:
```yaml
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: apicast-operator-metrics-reader
rules:
- nonResourceURLs:
  - /metrics
  verbs:
  - get
```

* In addition, Openshift monitoring team is considering a different way of securying metrics. Bear in mind that using bearer tokens for metrics authn puts additional load on the API server and they are looking at replacing this by client TLS auth in the future (it's being discussed here: [openshift/enhancements#701](https://github.com/openshift/enhancements/pull/701))

The downside is that the operator metrics could be read by anyone within the cluster. Two comments about this:
* apicast deployment (not the operator) metrics are currently, "open". Does not make much sense to have operator metrics secured and other metrics open.
* ServiceMonitor needs to be configured with TLS cert verification disabled. So, kube-rbac-proxy enforces security, but the connection is not verified. 

Having into account all this information, we have decided to remove the kube-rbac-proxy for now.


miguelsorianod (approved):
looks good to me

PR #179, opened 2022-10-11
bump version 0.7.0

eguzki:


MStokluska (approved):
Without understanding the process of releasing apicast operator fully I'm only able to verify if it makes sense and it looks good to merge to me

PR #198, opened 2023-07-07
Update controller-runtime to v0.15.0 & client-go to v0.28.0-alpha.2

austincunningham:
…double the requests and limits on the operator deployment


.circleci/config.yml
```diff
@@ -34,7 +34,7 @@ commands:
 jobs:
   install-operator:
     docker:
-      - image: cimg/go:1.18
```
eguzki, line 37: Is Golang 1.20 ready to be used in the productization pipeline?

austincunningham: hmm good question , I'm not sure , if not we will have to put this on hold for a while.

MStokluska:
@austincunningham is this still needed?

austincunningham:
@MStokluska no @carlkyrillos will have updated this

PR #207, opened 2024-01-11
THREESCALE-10366 initial config of HPA for Apicast Operator [WIP]

MStokluska:
Jira: https://issues.redhat.com/browse/THREESCALE-10366

# Verification

- osd/ocp cluster
- run against cluster:
```
oc new-project apicast-test
```
```
make install
```
```
make run
```
- on another namespace install 3scale operator (with APIM etc, either from local or operators hub)
- create a secret pointing to the 3scale admin portal to be consumed by the apicast CR on the same namespace where apicast operator is watching
```
kubectl create secret generic my-test --from-literal=AdminPortalURL=https://<ADMIN_TOKEN>@<ADMIN ROUTE - without https://>
```
- create apicast CR:
```
kind: APIcast
metadata:
  name: example-apicast
  namespace: apicast-test
spec:
  adminPortalCredentialsRef:
    name: my-test
  replicas: 4
  resources:
    limits:
      cpu: 2000m
      memory: 500Mi
    requests:
      cpu: 300m
      memory: 150Mi
```
- Wait for apicast CR to report completed
- The outcome here is that apicast pods will start up, with 4 replicas and above limits and request
- Set the `hpa: true` on apicast CR
- Outcome will be that the hpa object will be created, with 1-5 pods and initial value of pods matching the previous value which was - 4
- Let it run for about 10 minutes, the number of replicas should reduce over time
- Confirm that a warning condition is visible on the apicast CR indicating that requests, limits, and replicas values are ignored because HPA took over.
- Confirm that requests and limits are set to the same values 

```
	cpuLimits := resource.MustParse("1")
	memoryLimits := resource.MustParse("128Mi")
```
- Set the HPA to false / or remove it. The resource limits, requests and replicas should go back to old values (or the ones specified in the APICast CR) and the warning message should be removed from the CR.


MStokluska:
TODO: docs to be updated post final review

bundle/manifests/apicast-operator.clusterserviceversion.yaml
```diff
@@ -214,6 +214,17 @@ spec:
...
+          - autoscaling
+          resources:
+          - horizontalpodautoscaler
+          verbs:
+          - create
+          - delete
+          - get
+          - list
+          - update
+          - watch
```
austincunningham, line 227: been thinking about the premissions we really only need create and delete now.

controllers/apps/apicast_controller_status.go
```diff
@@ -95,6 +97,28 @@ func (r *APIcastReconciler) deploymentImage(ctx context.Context, cr *appsv1alpha
...
+		Message: "HorizontalPodAutoscaling (Hpa) enabled overrides values applied to request, limits and replicas",
+	}
+
+	// check if condition is already present
+	foundCondition := meta.FindStatusCondition(*conditions, "Warning")
+
+	// If hpa is enabled but the condition is not found, add it
+	if cr.Spec.Hpa && foundCondition == nil {
+		meta.SetStatusCondition(conditions, *cond)
+	}
```
austincunningham, line 114: hmm thinking can I work with this? it is fine for values in the cr.Spec, but say you are trying to satify some other logical condition not tied to the CR like say a value in a secret, Are we going to have to bring those values into the status reconciler to add a message :thinking:

MStokluska: I think HPA is unique since it's either on/off on the cr itself. So it's fine like that imo, but for more advanced warning we would need to make more changes

austincunningham:
## Verification
- [x] Apicast created with 4 replicas
- [x] Deployment had the specified requests and limits
- [x] Set `spec.hpa: true` and the deployment requests == limits and after about 10 min the replicas scale down
```bash
oc get hpa
NAME              REFERENCE                            TARGETS           MINPODS   MAXPODS   REPLICAS   AGE
example-apicast   Deployment/apicast-example-apicast   23%/90%, 0%/90%   1         5         1          10m
```
Weird issue in the ui seems to always have one pod pending
image

- [x] confirm the message in the CR
```yaml
status:
  conditions:
    - lastTransitionTime: '2024-01-18T14:28:53Z'
      message: APIcast is ready
      reason: Ready
      status: 'True'
      type: Ready
    - lastTransitionTime: '2024-01-18T14:28:53Z'
      message: >-
        HorizontalPodAutoscaling (Hpa) enabled overrides values applied to
        request, limits and replicas
      reason: HPA
      status: 'True'
      type: Warning
  image: 'quay.io/3scale/apicast:latest'
  observedGeneration: 5
```
- [x] remove `spec.hpa: true` and the pods scale up and deployment requests and limits match those specified in the CR
>NOTE: pending pods go away, but will appear again once hpa in enabled, don't believe its related to this PR . Wonder if its an issue with the HPA controller 
image

austincunningham (approved):
/lgtm

doc/operator-user-guide.md
```diff
@@ -169,6 +170,76 @@ spec:
...
+kind: APIcast
+metadata:
+  name: example-apicast
+  namespace: apicast
+spec:
+  adminPortalCredentialsRef:
+    name: <Admin portal credentials reference>
+  hpa:
+    enabled: true
+    minPods: 1
```
austincunningham, line 236: guess we need this updated first before we can merge.

doc/operator-user-guide.md
```diff
@@ -169,6 +170,59 @@ spec:
...
 Details about the available fields in the `exposedHost` section can be found [here](apicast-crd-reference.md#APIcastExposedHost)
 
+#### Setting Horizontal Pod Autoscaling 
+Horizontal Pod Autoscaling(HPA) is available for Apicasts. To enable HPA set the apicast.spec.hpa to `true`. HPA will be created with default values.
+
+- replicas: min 1; max 5;
+- request resource requirements: cpu: 1000m; memory: 128Mi;
+- limits resource requirements: cpu: 1000m; memory: 128Mi;
+
+Requests and limits values can be overwriten within the deployment itself.
```
austincunningham, line 180: Don't think we need this line.

austincunningham (approved):

PR #208, opened 2024-03-07
THREESCALE-10869 Bump Go, controller-runtime, and k8s dependencies

carlkyrillos:
# Issue Link
JIRA: [THREESCALE-10869](https://issues.redhat.com/browse/THREESCALE-10869)

# What
This PR bumps a number of dependencies, the notable ones being:
- `go:                 1.19   -> 1.20`
- `ginkgo:             1.16.5 -> 2.13.0`
- `client-go:          0.24.3 -> 0.28.3`
- `apimachinery:       0.24.3 -> 0.28.3`
- `controller-runtime: 0.12.2 -> 0.16.5`

NOTE: This PR doesn't bump the version of `operator-sdk`, that will be done in a separate PR.

# Verification Steps
## Prerequisites
1. Go v1.20 - I recommend using the [gvm](https://github.com/moovweb/gvm) CLI to quickly switch between Go versions
2. Provisioned OSD cluster

## Cluster Scoped Scenario
1. Create the CatalogSource and install the operator (note: feel free to [build](https://github.com/3scale/apicast-operator/blob/master/doc/development.md#deploy-custom-apicast-operator-using-olm) your own images or use the ones provided)
```bash
make bundle-run quay.io/ckyrillo/apicast-operator-bundles:threescale-10683
```

2. Create a Namespace, config Secret, and APIcast CR:
```bash
export NAMESPACE=apicast-test
oc new-project $NAMESPACE

cat << EOF | oc create -f -
apiVersion: v1
kind: Secret
metadata:
  name: apicast-config-secret
  namespace: $NAMESPACE
type: Opaque
stringData:
  config.json: |
    {
      "services": [
        {
          "proxy": {
            "policy_chain": [
                "configuration": {
                  "rules": [{
                    "regex": "/",
                  }]
                }
              }
            ]
          }
        }
      ]
    }
EOF

cat << EOF | oc create -f -
kind: APIcast
metadata:
  name: example-apicast
  namespace: $NAMESPACE
spec:
    name: apicast-config-secret
EOF
```

3. Verify the installation completes successfully:
```bash
oc get apicast -n $NAMESPACE -o yaml -w
```

4. Verify that the APIcast gateway is running and available
* First, port-forward the apicast service:
```bash
oc port-forward svc/apicast-example-apicast 8080
```
* Then curl the endpoint in a separate terminal and confirm you get a response:
```bash
curl 127.0.0.1:8080/test -H "Host: localhost"
```

5. Delete the test namespace to prepare for the namespace scoped scenario
```bash
oc delete project $NAMESPACE
```

6. Uninstall the cluster scoped operator to prepare for the namespace scoped scenario

## Namespace Scoped Scenario
1. Create a fresh Namespace, config Secret, and APIcast CR:
```bash
export NAMESPACE=apicast-test
oc new-project $NAMESPACE

cat << EOF | oc create -f -
apiVersion: v1
kind: Secret
metadata:
  name: apicast-config-secret
  namespace: $NAMESPACE
type: Opaque
stringData:
  config.json: |
    {
      "services": [
        {
          "proxy": {
            "policy_chain": [
                "configuration": {
                  "rules": [{
                    "regex": "/",
                  }]
                }
              }
            ]
          }
        }
      ]
    }
EOF

cat << EOF | oc create -f -
kind: APIcast
metadata:
  name: example-apicast
  namespace: $NAMESPACE
spec:
    name: apicast-config-secret
EOF
```
2. Navigate to the `OperatorHub` in the OpenShift UI, search for `apicast`, and install the dev version the apicast-operator in the `apicast-test` Namespace.

3. Verify the installation completes successfully:
```bash
oc get apicast -n $NAMESPACE -o yaml -w
```

4. Verify that the APIcast gateway is running and available
* First, port-forward the apicast service:
```bash
oc port-forward svc/apicast-example-apicast 8080
```
* Then curl the endpoint in a separate terminal and confirm you get a response:
```bash
curl 127.0.0.1:8080/test -H "Host: localhost"
```


Makefile
```diff
@@ -184,7 +184,7 @@ test-unit: generate fmt vet manifests ## Run Unit tests.
 # 3) Timeout waiting for process kube-apiserver to stop https://github.com/kubernetes-sigs/controller-runtime/issues/1571
 test-integration: export USE_EXISTING_CLUSTER=true
 test-integration: generate fmt vet manifests ## Run Integration tests.
-	go test ./... -tags integration -ginkgo.v -ginkgo.progress -v -timeout 600s
```
eguzki, line 187: consider using ginkgo CLI

I changed it recently in another operator https://github.com/Kuadrant/kuadrant-operator/pull/425/commits/f3553fa525d3e1b84c39f8533eecbdf22b5716a6

controllers/apps/apicast_controller.go
```diff
@@ -142,8 +140,13 @@ func (r *APIcastReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&appsv1alpha1.APIcast{}).
 		Watches(
-			&source.Kind{Type: &v1.Secret{}},
-			handler.EnqueueRequestsFromMapFunc(secretToApicastEventMapper.Map),
+			&corev1.Secret{},
+			handler.EnqueueRequestsFromMapFunc(
+				func(ctx context.Context, obj client.Object) []reconcile.Request {
+					secret := &corev1.Secret{}
+					reqs := secretToApicastEventMapper.Map(secret)
```
eguzki, line 147: this does not look good. Unless I am missing something, the `secretToApicastEventMapper.Map` function needs the `obj` object. Mainly because it needs the UID and that secret is empty.

controllers/apps/suite_test.go
```diff
@@ -49,53 +48,55 @@ func testClient() client.Client { return k8sClient }
...
+		Expect(err).ToNot(HaveOccurred())
+		err = (&APIcastReconciler{
+			BaseControllerReconciler: reconcilers.NewBaseControllerReconciler(mgr.GetClient(), mgr.GetAPIReader(), mgr.GetScheme()),
+			Log:                      ctrl.Log.WithName("controllers").WithName("APIcast"),
+		}).SetupWithManager(mgr)
 		Expect(err).ToNot(HaveOccurred())
-	}()
 
-}, 60)
+		go func() {
```
eguzki, line 90: another goroutine? really needed?

main.go
```diff
@@ -76,17 +79,38 @@ func main() {
...
-		LeaderElectionID:   "988b4062.3scale.net",
-	})
-	if err != nil {
-		setupLog.Error(err, "unable init the manager")
-		os.Exit(1)
+	var mgr ctrl.Manager
+	// If a watch namespace is detected (i.e. operator is namespace scoped), then pass the NS to cache.Options.DefaultNamespaces
+	// If no watch namespace is detected (i.e. operator is cluster scoped), then don't specify cache.Options.DefaultNamespaces
+	if namespace != "" {
+		mgr, err = ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
```
eguzki, line 86: what about one single `mgr, err = ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{...}`

And have `Cache` object with some value or another depending of the namespace scoping?

eguzki (changes requested):
overall looks good, some changes requested but I could be wrong

eguzki:
My main concern is the upgrade of controller-runtime beyond v0.15.0 as v0.15.0 introduced large list of breaking changed. I review them and nothing caught my eye. But second pair of :eye: would be very nice https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.15.0

controllers/apps/suite_test.go
carlkyrillos: The bump to ginkgo v2 changed the function signature and we can no longer pass in the `60` timeout using the old method. For what it's worth, I've found the test to be much faster and better with teardown using the goroutine.

main.go
carlkyrillos: I'll look into this but I think the issue we'll run into is we'll be passing in a nil `Cache` object when it's cluster scoped which could cause problems. The `Cache.DefaultNamespaces` field needs to be unset for cluster scoped.

carlkyrillos:
> My main concern is the upgrade of controller-runtime beyond v0.15.0 as v0.15.0 introduced large list of breaking changed. I review them and nothing caught my eye. But second pair of 👁️ would be very nice https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.15.0

@eguzki I [reviewed](https://github.com/kubernetes-sigs/controller-runtime/compare/v0.12.2...v0.16.5) every breaking change for every version from v0.12.2 to v0.16.5 and cross referenced them with the source code to ensure we weren't affected (or I included the fix to support the breaking change if we were affected). I did this when performing the [3scale-operator bump](https://github.com/3scale/3scale-operator/pull/936) as well. That being said, more eyes reviewing it is always a good thing.

main.go
carlkyrillos: Actually it looks like you can pass `cache.AllNamespaces` which would fix the nil `cache` object issue, I'll push this change as a new commit

Makefile
carlkyrillos: @eguzki What's the advantage of using the ginkgo CLI?

eguzki: Nothing really important that we need not. But reading the [doc](https://onsi.github.io/ginkgo/)

```
Under the hood, ginkgo is simply calling go test. While you can run go test instead of the ginkgo CLI, Ginkgo has several capabilities that can only be accessed via ginkgo. We generally recommend users embrace the ginkgo CLI and treat it as a first-class member of their testing toolchain.
```

controllers/apps/apicast_controller.go
eguzki: The 3scale operator implementation seems correct in this regard. 

You can try. The operator will add some labels with the UID of the secret to the APIcast CR. . Try updating the secret and see if that triggers a new reconciliation event for APIcast CR.

controllers/apps/suite_test.go
eguzki: What is the value provided? I might be missing something, but I do not think we need that sync mechanism you implemented with the channel. 

Check the `suite_test.go` in the [kuadrant operator](https://github.com/Kuadrant/kuadrant-operator/blob/main/controllers/suite_test.go). It's working well as far as I can tell.

Makefile
carlkyrillos: That's good to know thank you. Since this is more of a nice to have though, I'm going to just leave this as a follow up task because of the upcoming code freeze

controllers/apps/suite_test.go
carlkyrillos: @eguzki The value provided is that we preserve the existing behavior of the `60` timeout like we had with ginkgo v1. I'm also confused by the kuadrant-operator example because it looks like it also has that [goroutine](https://github.com/Kuadrant/kuadrant-operator/blob/main/controllers/suite_test.go#L230-L234) unless we're referring to different code blocks.

We've been using this implementation with [RHOAM](https://github.com/integr8ly/integreatly-operator/blob/master/test/functional/suite_test.go#L79-L106) for awhile now and haven't had any issues with it.

main.go
carlkyrillos: This is more for future readers but I was unable to use the `cache.AllNamespaces` string as intended. So I changed the implementation to still use one single `mgr, err = ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{...}` but instead have a `Cache` object which takes the namespace if it exists and otherwise is passed as an empty object.

controllers/apps/secret_to_apicast_event_mapper.go
```diff
@@ -13,12 +13,13 @@ import (
...
 // SecretToApicastEventMapper is an EventHandler that maps secret object to apicast CR's
 type SecretToApicastEventMapper struct {
+	Context   context.Context
 	K8sClient client.Client
 	Logger    logr.Logger
 	Namespace string
 }
 
-func (s *SecretToApicastEventMapper) Map(obj client.Object) []reconcile.Request {
+func (s *SecretToApicastEventMapper) Map(ctx context.Context, obj client.Object) []reconcile.Request {
```
eguzki, line 22: use the context!

controllers/apps/suite_test.go
eguzki: I am fine with this goroutine, it is well known. I was asking for the other one in the top and the sync mechanism to respect the 60 secs. 
```go
done := make(chan interface{})
go func() {
....
}
Eventually(done, 60).Should(BeClosed())
```

My understanding (corrrect me if I am wrong), is that this is enforcing the test setup (not the entire test) to happen within 60sec time period. The test setup being starting the test environment and instantiate the managers and controllers and finally, start the manager. That makes sense, when using the kubebuilder test environment, a kind of k8s fake control plane. However, we are using a real k8s control plane which is started before running the e2e tests. So, does it still makes sense? I am ok to keep that, not a big deal.

main.go
```diff
@@ -76,16 +79,27 @@ func main() {
...
-		Namespace:          namespace,
-		Scheme:             scheme,
-		MetricsBindAddress: metricsAddr,
-		Port:               9443,
-		LeaderElection:     enableLeaderElection,
-		LeaderElectionID:   "988b4062.3scale.net",
+		Cache:            managerCache,
+		Scheme:           scheme,
+		Metrics:          metricsserver.Options{BindAddress: metricsAddr},
+		WebhookServer:    webhook.NewServer(webhook.Options{Port: 9443}),
```
eguzki, line 97: It is unrelated, but do we use and feature for the webhook server? could we disable it?

eguzki:
this is getting pretty close to be merged. 

Just one iteration more with few comments and we should be ready to merge.

main.go
carlkyrillos: I'm provisioning a cluster now to confirm but I think you are correct and we can disable it

carlkyrillos: I confirmed we can remove it and pushed the changes

controllers/apps/secret_to_apicast_event_mapper.go
carlkyrillos: Fixed

controllers/apps/suite_test.go
carlkyrillos: Ah okay that makes sense thank you for the explanation. I fixed it and pushed the changes

eguzki (approved):
good job! :medal_sports:

PR #221, opened 2024-08-07
THREESCALE-11230 set version to 0.12.0

MStokluska:


valerymo:
Same version in 3scale-operator:
https://github.com/3scale/3scale-operator/blob/3scale-2.15-stable/version/version.go#L8
/lgtm

PR #233, opened 2024-10-11
THREESCALE-11396 Improved secret watched-by logic

carlkyrillos:
# Issue Link
JIRA: [THREESCALE-11396](https://issues.redhat.com/browse/THREESCALE-11396)

# What
This PR improves on the existing secret watched-by logic in a few ways:

1. Previously the secret labels on the APIcast CR followed this pattern: `secret.apicast.apps.3scale.net/{secret-UID: 'true'`, where the value was set to `true` regardless of whether the secret had the `watched-by` label. Now the value is set to `false` if the Secret doesn't have the `watched-by` label and `true` if the secret does have the `watched-by` label.
2. Previously the operator would annotate the apicast pod with `apimanager.apps.3scale.net/{secret-name}: '{secret-resourceVersion}'` whenever a secret was referenced in the APIcast CR. Now the apicast pod is only annotated if the referenced secret _also_ has the `watched-by` label on it.
3. Previously, any and all changes to a watched secret would trigger the apicast deployment to rollout a new pod whose annotations contained the latest resourceVersion of the watched secret. Now the operator will only rollout new pods when there is a change to the secret's `.data`, i.e. changes to the watched secrets labels, annotations, etc. won't trigger a rollout even if though the secret's resourceVersion changed. This is accomplished through a new secret called `hashed-secret-data` that stores a hash of each watched secret's data using SHA256 encryption. Whenever the operator detects that a watched secret's resourceVersion has changed, it first takes a hash of the secret's current `.data` and if that matches the secret's entry in the `hashed-secret-data` secret, then the operator will ignore the change and prevent a rollout. If the `.data` has changed, the operator will update the apicast pod's annotations with the watched secret's latest resourceVersion which will trigger a rollout.

# Verification Steps
1. Checkout this PR
2. Prepare the cluster for a local install:
```bash
make download
make install
```
3. Create a Namespace and the configuration Secret:
```bash
export NAMESPACE=apicast-test
oc new-project $NAMESPACE

cat << EOF | oc create -f -
apiVersion: v1
kind: Secret
metadata:
  name: apicast-config-secret
  namespace: $NAMESPACE
type: Opaque
stringData:
  config.json: |
    {
      "services": [
        {
          "proxy": {
            "policy_chain": [
                "configuration": {
                  "rules": [{
                    "regex": "/",
                  }]
                }
              }
            ]
          }
        }
      ]
    }
EOF
```

4. Create a custom environment Secret (without the watched-by label):
```bash
cat << EOF | oc create -f -
apiVersion: v1
kind: Secret
metadata:
  name: custom-env-1
  namespace: $NAMESPACE
type: Opaque
stringData:
  custom_env.lua: |
    local cjson = require('cjson')
    local PolicyChain = require('apicast.policy_chain')
    local policy_chain = context.policy_chain
    
    local logging_policy_config = cjson.decode([[
    {
      "enable_access_logs": false,
      "custom_logging": "\"{{request}}\" to service {{service.name}} and {{service.id}}"
    }
    ]])
    
    policy_chain:insert( PolicyChain.load_policy('logging', 'builtin', logging_policy_config), 1)
    
    return {
      policy_chain = policy_chain,
      port = { metrics = 9421 },
    }
EOF
```

5. Create an APIcast CR that references the `custom-env-1` Secret:
```bash
cat << EOF | oc create -f -
kind: APIcast
metadata:
  name: example-apicast
  namespace: $NAMESPACE
spec:
    name: apicast-config-secret
  customEnvironments:
    - secretRef:
        name: custom-env-1
EOF
```

6. Run the operator:
```bash
make run
```

7. Wait for APIcast to be ready:
```bash
oc get deployment apicast-example-apicast -oyaml | yq '.status'
```

8. Verify that the APIcast labels reference the two secrets' UIDs but the label values are `false`:
```bash
oc get apicast example-apicast -oyaml | yq '.metadata.labels'
```
The labels should look like this:
```
secret.apicast.apps.3scale.net/86655a7f-c4bd-4d2a-b364-9fcdfe8eed96: "false"
secret.apicast.apps.3scale.net/fdb8d92d-e446-4882-aa49-569329f5712d: "false"
```

9. Verify that the apicast pod annotations don't contain references to the secrets since neither of the secrets have the `watched-by` label:
```bash
oc get pods -o yaml | yq '.items[0].metadata.annotations' | grep apicast.apps.3scale.net
```

10. Verify that the `hashed-secret-data` secret exists but is empty:
```bash
oc get secret hashed-secret-data -oyaml
```

11. Add the `watched-by` label to the configuration secret:
```bash
oc label secret apicast-config-secret apicast.apps.3scale.net/watched-by=apicast
```

12. Verify that the APIcast labels were updated and that one of secret labels has a value of `true`:
```bash
oc get apicast example-apicast -oyaml | yq '.metadata.labels'
```
The labels should now look like this:
```
secret.apicast.apps.3scale.net/86655a7f-c4bd-4d2a-b364-9fcdfe8eed96: "false"
secret.apicast.apps.3scale.net/fdb8d92d-e446-4882-aa49-569329f5712d: "true"
```

13. Once the new pod is ready, verify that it has an annotation referencing the config secret:
```bash
oc get pods -o yaml | yq '.items[0].metadata.annotations' | grep apicast.apps.3scale.net
```
There should be an annotation that looks like this:
```
apicast.apps.3scale.net/gateway-configuration-secret-resource-version: "164701"
```

14. Verify that the `hashed-secret-data` secret has an entry for the config secret:
```bash
oc get secret hashed-secret-data -oyaml
```

15. Edit the data in the config secret and verify a new pod is created:
```bash
oc get pods
```

16. Once the pods have stabilized, add a label to the config secret and verify that no new pods are created even though the resourceVersion has changed:
```bash
oc label secret apicast-config-secret dummy=label
```

### DIFF
--- a/apis/apps/v1alpha1/apicast_types.go
+++ b/apis/apps/v1alpha1/apicast_types.go
@@ -466,3 +466,91 @@ func (a *APIcast) Validate() field.ErrorList {
 func init() {
 	SchemeBuilder.Register(&APIcast{}, &APIcastList{})
 }
+
+func (a *APIcast) GetAdminPortalCredentialsSecretRef() *v1.LocalObjectReference {
+	return a.Spec.AdminPortalCredentialsRef
+}
+
+func (a *APIcast) GetEmbeddedConfigurationSecretRef() *v1.LocalObjectReference {
+	return a.Spec.EmbeddedConfigurationSecretRef
+}
+
+func (a *APIcast) GetOpenTelemetrySecretRef() *v1.LocalObjectReference {
+	if a.Spec.OpenTelemetry == nil {
+		return nil
+	}
+	return a.Spec.OpenTelemetry.TracingConfigSecretRef
+}
+
+func (a *APIcast) GetOpenTracingSecretRef() *v1.LocalObjectReference {
+	if a.Spec.OpenTracing == nil {
+		return nil
+	}
+	return a.Spec.OpenTracing.TracingConfigSecretRef
+}
+
+func (a *APIcast) GetCustomEnvironmentsSecretRefs() []*v1.LocalObjectReference {
+	if a.Spec.CustomEnvironments == nil {
+		return nil
+	}
+
+	secretRefs := []*v1.LocalObjectReference{}
+	for _, env := range a.Spec.CustomEnvironments {
+		if env.SecretRef != nil {
+			secretRefs = append(secretRefs, env.SecretRef)
+		}
+	}
+
+	return secretRefs
+}
+
+func (a *APIcast) GetCustomPoliciesSecretRefs() []*v1.LocalObjectReference {
+	if a.Spec.CustomPolicies == nil {
+		return nil
+	}
+
+	secretRefs := []*v1.LocalObjectReference{}
+	for _, policy := range a.Spec.CustomPolicies {
+		if policy.SecretRef != nil {
+			secretRefs = append(secretRefs, policy.SecretRef)
+		}
+	}
+
+	return secretRefs
+}
+
+func (a *APIcast) GetApicastSecretRefs() []*v1.LocalObjectReference {
+	secretRefs := []*v1.LocalObjectReference{}
+
+	adminPortalCredentialsSecretRef := a.GetAdminPortalCredentialsSecretRef()
+	if adminPortalCredentialsSecretRef != nil {
+		secretRefs = append(secretRefs, adminPortalCredentialsSecretRef)
+	}
+
+	embeddedConfigurationSecretRef := a.GetEmbeddedConfigurationSecretRef()
+	if embeddedConfigurationSecretRef != nil {
+		secretRefs = append(secretRefs, embeddedConfigurationSecretRef)
+	}
+
+	openTelemetrySecretRef := a.GetOpenTelemetrySecretRef()
+	if openTelemetrySecretRef != nil {
+		secretRefs = append(secretRefs, openTelemetrySecretRef)
+	}
+
+	opentracingSecretRef := a.GetOpenTracingSecretRef()
+	if opentracingSecretRef != nil {
+		secretRefs = append(secretRefs, opentracingSecretRef)
+	}
+
+	customEnvironmentSecretRefs := a.GetCustomEnvironmentsSecretRefs()
+	if customEnvironmentSecretRefs != nil && len(customEnvironmentSecretRefs) > 0 {
+		secretRefs = append(secretRefs, customEnvironmentSecretRefs...)
+	}
+
+	customPoliciesSecretRefs := a.GetCustomPoliciesSecretRefs()
+	if customPoliciesSecretRefs != nil && len(customPoliciesSecretRefs) > 0 {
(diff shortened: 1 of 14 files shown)


valerymo:
hey @carlkyrillos 
I verified PR. Done very good, working as described in verification steps. Minor comments in code - up to you.

But I have Comments for design. Can you please look.
- Current approach is much stricter than the approach in 3scale PR https://github.com/3scale/3scale-operator/pull/861,
where any change in the secret referenced in the CR (APImanager) by the label apimanager.apps.3scale.net/watched-by=apimanager triggers a pod re-creation.  
- In the current solution - only a change in secret DATA triggers a Pod(s) re-creation.  
      -  all information used for the operator logic will have to be in secret Data (?) 
      - label and annotations in Secrets are useless (?)
- Can it conflict with the existing design/functionality ( that not in this PR). Was it analized, or maybe discussed with Eguzki?
- I think we need to open a separate task for regression testing.

Thanks

valerymo:
@carlkyrillos , seems I was wrong when thaought that Labels and annotations become useless in Secrets that received "watched-by" label/. 
Metadata still will be used in logics, just changes in it will not cause Pod(s) restarting. But still maybe need think more about this + Regression.

austincunningham (approved):
/approve

apis/apps/v1alpha1/apicast_types.go
```diff
@@ -466,3 +466,91 @@ func (a *APIcast) Validate() field.ErrorList {
...
+}
+
+func (a *APIcast) GetOpenTelemetrySecretRef() *v1.LocalObjectReference {
+	if a.Spec.OpenTelemetry == nil {
+		return nil
+	}
+	return a.Spec.OpenTelemetry.TracingConfigSecretRef
+}
+
+func (a *APIcast) GetOpenTracingSecretRef() *v1.LocalObjectReference {
```
MStokluska, line 485: this will be dropped in 2.16 so feel free to skip any openTracing stuff

pkg/apicast/apicast.go
```diff
@@ -422,16 +434,119 @@ func (a *APIcast) Deployment() *appsv1.Deployment {
...
+	return uncheckedAnnotations, nil // No difference with existing annotations so can return uncheckedAnnotations
+}
+
+func (a *APIcast) getWatchedSecretAnnotations() map[string]string {
+	// https certificate secret
+	// admin portal secret
+	// gateway conf secret
+	// custom policy secret(s)
+	// custom env secret(s)
+	// tracing config secret
```
MStokluska, line 502: out of interest, why is tracing here but telemetry isn't?

apis/apps/v1alpha1/apicast_types.go
carlkyrillos: If there's no harm in keeping it then I might just leave it in. But let me know if it needs to be removed.

MStokluska (approved):
looks good to me. Might be a good idea to document which files need to be changed to support watched-by on new secrets, but up to you.

MStokluska (approved):

PR #234, opened 2024-11-01
THREESCALE-11396 Fixed hashing multi-key secrets

carlkyrillos:
# Issue Link
JIRA: [THREESCALE-11396](https://issues.redhat.com/browse/THREESCALE-11396)

# What
This PR fixes the hash implementation when hashing multi-key secrets.

# Verification Steps
Passing the PROW checks and an eye review should be sufficient verification.

### DIFF
--- a/pkg/apicast/apicast.go
+++ b/pkg/apicast/apicast.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"path"
 	"reflect"
+	"sort"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


pkg/apicast/apicast.go
```diff
@@ -740,7 +741,14 @@ func (a *APIcast) computeHashedSecretData(ctx context.Context, k8sclient client.
...
+	for k := range data {
+		sortedKeys = append(sortedKeys, k)
+	}
+	sort.Strings(sortedKeys)
+
+	for _, key := range sortedKeys {
+		value := data[key]
 		combinedKeyValue := append([]byte(key), value...)
 		hash.Write(combinedKeyValue)
 	}
```
valerymo, line 754: @carlkyrillos It looks good, but I'd like to ask - what is the reason that key and value are treated as a single entity when hashing?
Another approach - distinct entities:
for _, key := range sortedKeys {
	hash.Write([]byte(key))
	hash.Write(data[key])
}
Thanks

valerymo (commented):
Looks good, just quesion - why key and value are treated as a single entity when hashing? (please see comment for other example)
Thanks

pkg/apicast/apicast.go
carlkyrillos: I combined the key and value so that the hash will change if either the key _or_ the value changes because either should trigger an update.

austincunningham (approved):
/approve

pkg/apicast/apicast.go
valerymo: hey @carlkyrillos , seems to me that in my example hash will also change if either the key or the value changes. But doesn't matter, if I will have time, I will check, but later. Thank you

PR #258, opened 2025-07-18
THREESCALE-9972 - Add IngressClassName field in the Ingress object created by the APIcast operator

tkan145:
## What

Fix https://issues.redhat.com/browse/THREESCALE-9972

TLDR;
* The OpenShift Ingress Controller team decided to add this warning in OpenShift 4.12, but many ingress controllers still offer the "default controller" feature (see [this comparison page](https://hackmd.io/@maelvls/k8s-ingress-controllers-impl-comparison)), meaning that it is possible to use an Ingress resource without setting the annotation or ingressClassName

I found an explanation as to why this warning exists in [transition-ingress-from-beta-to-stable.md](https://github.com/openshift/enhancements/blob/master/enhancements/ingress/transition-ingress-from-beta-to-stable.md#risks-and-mitigations):
> We are considering adding alerts in case any Routes existed in this state, so that the administrator would know that the Routes needed to be deleted, or the Ingress modified to specify an appropriate IngressClass so that OpenShift would once again reconcile the Routes.

* An OpenShift customer, says that they are spammed by a lot of alerts since they upgraded to OpenShift 4.12

## Verification steps
* Checkout this branch
* Run
```
make install
```
* Create Apicast config
```
export NAMESPACE=apicast-test
oc new-project $NAMESPACE

cat << EOF | oc create -f -
apiVersion: v1
kind: Secret
metadata:
  name: apicast-config-secret
  namespace: $NAMESPACE
type: Opaque
stringData:
  config.json: |
    {
      "services": [
        {
          "proxy": {
            "policy_chain": [
                "configuration": {
                  "rules": [{
                    "regex": "/",
                  }]
                }
              }
            ]
          }
        }
      ]
    }
EOF
```
* Create APIcast
```
cat << EOF | oc create -f -
kind: APIcast
metadata:
  name: example-apicast
  namespace: $NAMESPACE
spec:
  exposedHost:
    host: my-gateway
    ingressClassName: openshift-default
    name: apicast-config-secret
EOF
```
* Run operator locally
```
make run
```
* Wait APIcast pod comes online
* Check ingress object
```
 ▲ oc get ingress                                                            
NAME                      CLASS               HOSTS        ADDRESS                           PORTS   AGE 
apicast-example-apicast   openshift-default   my-gateway   router-default.apps-crc.testing   80      178m

 ▲ oc get ingress apicast-example-apicast -o json | jq -r .spec
{
  "ingressClassName": "openshift-default",
  "rules": [
    {
      "host": "my-gateway",
      "http": {
        "paths": [
          {
            "backend": {
              "service": {
                "name": "apicast-example-apicast",
                "port": {
                  "name": "proxy"
                }
              }
            },
            "pathType": "ImplementationSpecific"
          }
        ]
      }
    }
  ]
}
```
* Remove `ingressClassName` from APIcastCR
* Check Ingress object again, `ingressClassName` should be removed from the Ingress
```
 ▲ oc get ingress apicast-example-apicast -o json | jq -r .spec    
{
  "ingressClassName": "openshift-default",
  "rules": [
    {
      "host": "my-gateway",
      "http": {
        "paths": [
          {
            "backend": {
              "service": {
                "name": "apicast-example-apicast",
                "port": {
                  "name": "proxy"
                }
              }
            },
            "pathType": "ImplementationSpecific"
          }
        ]
      }
    }
  ]
}
```


apis/apps/v1alpha1/apicast_types.go
```diff
@@ -252,6 +252,8 @@ const (
 type APIcastExposedHost struct {
 	Host string `json:"host"`
 	// +optional
+	IngressClassName string `json:"ingressClassName"`
```
eguzki, line 255:
```suggestion
	IngressClassName string `json:"ingressClassName,omitempty"`
```

eguzki, line 255: consider having a pointer to `string`, but both options LGTM

pkg/reconcilers/ingress.go
```diff
@@ -29,6 +29,11 @@ func IngressMutator(existingObj, desiredObj k8sutils.KubernetesObject) (bool, er
 
 	update := false
 
+	if existing.Spec.IngressClassName != desired.Spec.IngressClassName {
```
eguzki, line 32: Wondering if pointer to structures should be compared with `reflect.DeepEqual(x,y)`?? otherwise it is very likely to be different even if the structure data is the same. makes sense?

tkan145: Yes you are right, they would point to different memory address.

tkan145: fixed

PR #261, opened 2025-09-11
THREESCALE-11930 Add support for pod priority

tkan145:
## What

[THREESCALE-11930](https://issues.redhat.com/browse/THREESCALE-11930)

## Verification steps
* Checkout this branch
* Run
```bash
make install
```
* Create Apicast config
```bash
export NAMESPACE=apicast-test
oc new-project $NAMESPACE

cat << EOF | oc create -f -
apiVersion: v1
kind: Secret
metadata:
  name: apicast-config-secret
  namespace: $NAMESPACE
type: Opaque
stringData:
  config.json: |
    {
      "services": [
        {
          "proxy": {
            "policy_chain": [
                "configuration": {
                  "rules": [{
                    "regex": "/",
                  }]
                }
              }
            ]
          }
        }
      ]
    }
EOF
```
* Create PriorityClass
```bash
cat << EOF | oc create -f -
apiVersion: scheduling.k8s.io/v1
kind: PriorityClass
metadata:
  name: high-priority
value: 1000000
globalDefault: false
description: "This priority class should be used for XYZ service pods only."
EOF
```
* Create APIcast
```bash
cat << EOF | oc create -f -
---
kind: APIcast
metadata:
  name: example-apicast
  namespace: $NAMESPACE
spec:
  exposedHost:
    host: my-gateway
    ingressClassName: openshift-default
    name: apicast-config-secret
  priorityClassName: high-priority
EOF
```
* Run operator locally
```
make run
```
* Check APIcast deployment, you should see the PriorityClassName applied
* Update priorityClassName to non-exist value, i.e `low-priority`, you shall see the following error in the status
```
pods "apicast-example-apicast-76c6b9bcb9-" is forbidden: no PriorityClass with name low-priority was found
```


eguzki (approved):
LGTM

verification steps working like a charm. 

Nitpicking here: it would be nicer to show the error about no priorityclass found in the status of the Apicast CR. That brings a more interesting idea out of the scope of this PR, maybe the APicast CR status should check the status of the deployment.

tkan145:
yeah agree, I've filled another issue and will revisit this idea when I have more bandwidth.

PR #271, opened 2026-02-25
THREESCALE-12224 Loosen APIcast operator logic to allow custom environment variables in Deployment

tkan145:
## What

Fix [THREESCALE-12224](https://issues.redhat.com/browse/THREESCALE-12224)

## Verification steps
* run `make install`
* run:
```
export NAMESPACE=apicast-test
oc new-project $NAMESPACE

cat << EOF | oc create -f -
apiVersion: v1
kind: Secret
metadata:
  name: apicast-config-secret
  namespace: $NAMESPACE
type: Opaque
stringData:
  config.json: |
    {
      "services": [
        {
          "proxy": {
            "policy_chain": [
                "configuration": {
                  "rules": [{
                    "regex": "/",
                  }]
                }
              }
            ]
          }
        }
      ]
    }
EOF
```
* run
```
cat << EOF | oc create -f -
kind: APIcast
metadata:
  name: example-apicast
  namespace: $NAMESPACE
spec:
    name: apicast-config-secret
EOF
```
 * run `make run`
 * Update Deployment en var
```
 oc set env deployment/apicast-example-apicast Foo=bar
```
* Confirm that new pod is deployed and check the pod variable environments
* Restart the operator
```
CTRL-C
make run
```
* Confirm that the environment variable is still there.
* Remove the environment variable from Deployment and check that it has also been removed from the pod.

### DIFF
--- a/pkg/reconcilers/deployment.go
+++ b/pkg/reconcilers/deployment.go
@@ -79,7 +79,7 @@ func DeploymentServiceAccountNameMutator(desired, existing *appsv1.Deployment) b
 }
 
 func DeploymentEnvVarsMutator(desired, existing *appsv1.Deployment) bool {
-	return ReconcileEnvVar(&existing.Spec.Template.Spec.Containers[0].Env, desired.Spec.Template.Spec.Containers[0].Env)
+	return ReconcileEnvVars(&existing.Spec.Template.Spec.Containers[0].Env, desired.Spec.Template.Spec.Containers[0].Env)
 }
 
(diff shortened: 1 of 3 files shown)


urbanikb (approved):
Tested in the cluster, operator no longer removes environment vars that customers add to the spec. Tests look great 👍 

Tried also with `oc set env deployment/apicast-example-apicast THREESCALE_CONFIG_FILE-` and the value has been reconciled back without removing the env var.

PR #274, opened 2026-04-21
Loosen APIcast operator logic to allow custom environment variables i…

tkan145:
## What

Backport #271 to 2.16


urbanikb:
Looks like integration test passed in the controllers package - which should be sufficient, but then errored out on the reconcilers package.

For the test, could you also backport this single commit: https://github.com/3scale/apicast-operator/commit/e5f4ae8f794c12b5d1462f7f7af130413d253f18

borisurbanik (approved):

PR #278, opened 2026-07-07
THREESCALE-15463 Bump k8s to v1.35.0 and controller-runtime to v0.23.3

tkan145:
## What
https://redhat.atlassian.net/browse/THREESCALE-15463

* Bumped k8s version to v1.35.4
* Bumped controller-runtime to v0.23.3 (latest against kubernetes v1.35.x)

### Breaking changes
* Fake client now strips the TypeMeta fields ⚠️ Fakeclient: Clear typemeta for structured  kubernetes-sigs/controller-runtime#3229.
Therefore, it is required to adjust how the test
* reconcile.Result.Requeue (bool) has been marked as deprecated

**NOTE**: This PR DID NOT fix reconcile.Result.Requeue (bool) deprecation to preserve reconciler behavior. Instead, this PR adds an exclusion rule for golangci-lint.

## Verification steps
* Create catalog source
```
cat << EOF | oc create -f -
apiVersion: operators.coreos.com/v1alpha1
kind: CatalogSource
metadata:
  name: threescale-dev
  namespace: openshift-marketplace
spec:
  sourceType: grpc
  image: quay.io/an_tran/apicast-operator-index:v0.17.98
EOF
```
* Wait for the catalog source to be install then head to the catalog to install APIcast operator from our custom catalog, pick version v2.16.4
* Once installed, run the following to setup APIcast
```
export NAMESPACE=apicast-test
oc new-project $NAMESPACE

cat << EOF | oc create -f -
apiVersion: v1
kind: Secret
metadata:
  name: apicast-config-secret
  namespace: $NAMESPACE
type: Opaque
stringData:
  config.json: |
    {
      "services": [
        {
          "proxy": {
            "policy_chain": [
                "configuration": {
                  "rules": [{
                    "regex": "/",
                  }]
                }
              }
            ]
          }
        }
      ]
    }
```

```
cat << EOF | oc create -f -
kind: APIcast
metadata:
  name: example-apicast
  namespace: $NAMESPACE
spec:
    name: apicast-config-secret
EOF
```
* Wait for the pod to be ready, then head to the subscription and upgrade to alpha channel
* You should be able to upgrade without any issue.


borisurbanik (approved):
I've gone through breaking changes in the changelogs and most of them we don't use (for example we don't use events, webhooks, or all the typed/generics features). 

Tested the upgrade path as described and there were no problems (the instructions are missing the CR creation, tested with a single field for the secret name (embeddedConfigurationSecretRef) and that was sufficient to create the deployment).

